PR #1, opened 2021-06-02
Basic UI features of home page and extraction form

SenecaAnderson:
Summary
his includes the code for the home page and the extractor form. The user can navigate between the two pages. The form can't be filled out or submitted. The inputs for the configuration file path and the log file path are currently text inputs instead of file inputs.

Code changes
Files for various components were created. Home.js and Extract.js have classes for the pages that will display. Stylesheets were created and modified.

Testing guidance
Start the app. When you click "Extract New", the extraction form will appear. Click "Cancel" to return to the home page.
npm --start

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,9 @@
     }
   },
   "dependencies": {
-    "electron-squirrel-startup": "^1.0.0"
+    "electron-squirrel-startup": "^1.0.0",
+    "react-router-bootstrap": "^0.25.0",
+    "react-router-dom": "^5.2.0"
   },
   "devDependencies": {
     "@electron-forge/cli": "^6.0.0-beta.54",
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     }
   },
   "dependencies": {
-    "electron-squirrel-startup": "^1.0.0",
+    "electron-squirrel-startup": "^1.0.0"
   },
   "devDependencies": {
     "@electron-forge/cli": "^6.0.0-beta.54",
--- a/package.json
+++ b/package.json
@@ -52,8 +52,6 @@
   },
   "dependencies": {
     "electron-squirrel-startup": "^1.0.0",
-    "react-router-bootstrap": "^0.25.0",
-    "react-router-dom": "^5.2.0"
   },
   "devDependencies": {
     "@electron-forge/cli": "^6.0.0-beta.54",
--- a/react-app/public/index.html
+++ b/react-app/public/index.html
@@ -10,6 +10,7 @@
       name="description"
       content="Web site created using create-react-app"
     />
+    <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.1.3/css/bootstrap.min.css" integrity="sha384-MCw98/SFnGE8fJT3GXwEOngsV7Zt27NXFoaoApmYm81iuXoPkFOJwJ8ERdknLPMO" crossorigin="anonymous">
     <link rel="apple-touch-icon" href="%PUBLIC_URL%/logo192.png" />
     <!--
       manifest.json provides metadata used when your web app is installed on a
@@ -40,5 +41,8 @@
       To begin the development, run `npm start` or `yarn start`.
       To create a production bundle, use `npm run build` or `yarn build`.
     -->
+    <script src="https://code.jquery.com/jquery-3.3.1.slim.min.js" integrity="sha384-q8i/X+965DzO0rT7abK41JStQIAqVgRVzpbzo5smXKp4YfRvH+8abtTE1Pi6jizo" crossorigin="anonymous"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.14.3/umd/popper.min.js" integrity="sha384-ZMP7rVo3mIykV+2+9J3UJ46jBk0WLaUAdn689aCwoqbBJiSnjAK/l8WvCWPIPm49" crossorigin="anonymous"></script>
+    <script src="https://stackpath.bootstrapcdn.com/bootstrap/4.1.3/js/bootstrap.min.js" integrity="sha384-ChfqqxuZUCnJSK3+MXmPNIyE6ZbWh2IMqE241rYiqJxyMiZ6OW/JmZQ5stwEULTy" crossorigin="anonymous"></script>
   </body>
 </html>
--- a/react-app/public/index.html
+++ b/react-app/public/index.html
@@ -40,8 +40,5 @@
       To begin the development, run `npm start` or `yarn start`.
       To create a production bundle, use `npm run build` or `yarn build`.
     -->
-    <script src="https://code.jquery.com/jquery-3.3.1.slim.min.js" integrity="sha384-q8i/X+965DzO0rT7abK41JStQIAqVgRVzpbzo5smXKp4YfRvH+8abtTE1Pi6jizo" crossorigin="anonymous"></script>
-    <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.14.3/umd/popper.min.js" integrity="sha384-ZMP7rVo3mIykV+2+9J3UJ46jBk0WLaUAdn689aCwoqbBJiSnjAK/l8WvCWPIPm49" crossorigin="anonymous"></script>
-    <script src="https://stackpath.bootstrapcdn.com/bootstrap/4.1.3/js/bootstrap.min.js" integrity="sha384-ChfqqxuZUCnJSK3+MXmPNIyE6ZbWh2IMqE241rYiqJxyMiZ6OW/JmZQ5stwEULTy" crossorigin="anonymous"></script>
   </body>
 </html>
--- a/react-app/public/index.html
+++ b/react-app/public/index.html
@@ -10,7 +10,6 @@
       name="description"
       content="Web site created using create-react-app"
     />
-    <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.1.3/css/bootstrap.min.css" integrity="sha384-MCw98/SFnGE8fJT3GXwEOngsV7Zt27NXFoaoApmYm81iuXoPkFOJwJ8ERdknLPMO" crossorigin="anonymous">
     <link rel="apple-touch-icon" href="%PUBLIC_URL%/logo192.png" />
     <!--
       manifest.json provides metadata used when your web app is installed on a
--- a/react-app/src/App.js
+++ b/react-app/src/App.js
@@ -1,56 +1,32 @@
-import React from "react";
-import {
-  BrowserRouter as Router,
-  Switch,
-  Route,
-} from "react-router-dom";
+import React from 'react';
+import { BrowserRouter as Router, Switch, Route } from 'react-router-dom';
 
-import Home from './components/Home'
-import Extract from './components/Extract'
-import Results from './components/Results'
+import Home from './components/Home';
+import Extract from './components/Extract';
 
 import 'bootstrap/dist/css/bootstrap.min.css';
 import './App.css';
 import './stylesheets/Page.css';
 
-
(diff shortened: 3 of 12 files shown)


react-app/public/index.html
```diff
@@ -10,6 +10,7 @@
       name="description"
       content="Web site created using create-react-app"
     />
+    <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.1.3/css/bootstrap.min.css" integrity="sha384-MCw98/SFnGE8fJT3GXwEOngsV7Zt27NXFoaoApmYm81iuXoPkFOJwJ8ERdknLPMO" crossorigin="anonymous">
```
mgramigna, line 13:
```suggestion
```

We should not need to use CDNs for CSS or JS content in `index.html`.

Since we set up the react app with `create-react-app`, we should be able to install these dependencies with just `npm install --save ___`

For bootstrap CSS specifically in this line, you can follow what the [React bootstrap docs](https://react-bootstrap.github.io/getting-started/introduction#installation) say for using `npm` to install the library

react-app/public/index.html
```diff
@@ -40,5 +41,8 @@
       To begin the development, run `npm start` or `yarn start`.
       To create a production bundle, use `npm run build` or `yarn build`.
     -->
+    <script src="https://code.jquery.com/jquery-3.3.1.slim.min.js" integrity="sha384-q8i/X+965DzO0rT7abK41JStQIAqVgRVzpbzo5smXKp4YfRvH+8abtTE1Pi6jizo" crossorigin="anonymous"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.14.3/umd/popper.min.js" integrity="sha384-ZMP7rVo3mIykV+2+9J3UJ46jBk0WLaUAdn689aCwoqbBJiSnjAK/l8WvCWPIPm49" crossorigin="anonymous"></script>
+    <script src="https://stackpath.bootstrapcdn.com/bootstrap/4.1.3/js/bootstrap.min.js" integrity="sha384-ChfqqxuZUCnJSK3+MXmPNIyE6ZbWh2IMqE241rYiqJxyMiZ6OW/JmZQ5stwEULTy" crossorigin="anonymous"></script>
```
mgramigna, line 46:
```suggestion
```

See above comment re: `npm install`

`jquery` and `popper` should be taken care of when you do your `npm install` for bootstrap, so we should not need to install those separately.

react-app/src/App.js
```diff
@@ -1,20 +1,56 @@
...
+      </div>
+    </Router>
+    </div>
+  );
+}
+
+export default App;
+
+
+/*
```
mgramigna, line 44: Can we get rid of the commented out code here that is no longer used?

react-app/src/App.js
```diff
@@ -1,20 +1,56 @@
...
+      <Router>
+      <div>
+        {/* A <Switch> looks through its children <Route>s and
+            renders the first one that matches the current URL. */}
+        <Switch>
+          <Route path="/extract">
+            <Extract />
+          </Route>
+          <Route path="/results">
+            <Results />
```
mgramigna, line 29: The app does not compile for me initially because this `Results` component is blank and has no exports, so React gets mad. Can we either delete this route for now (and it will later be added by your work to display the results), or just have `Results` be a simple component that renders just a `div` or something?

react-app/src/components/Extract.js
```diff
@@ -0,0 +1,106 @@
...
+            filterStart: false,
+            filterEnd: false
+        }
+
+        this.onConfigFileChange = this.onConfigFileChange.bind;
+        this.onLogFileChange = this.onLogFileChange.bind;
+        this.onIncludeDebugChange = this.onIncludeDebugChange.bind;
+        this.onFilterStartChange = this.onFilterStartChange.bind;
+        this.onFilterEndChange = this.onFilterEndChange.bind;
+        this.onSubmit = this.onSubmit.bind(this);
```
mgramigna, line 24:
```suggestion
```

The `.bind` motif is something that shouldn't be necessary in class components anymore thanks to arrow functions. I'll make some inline comments below showcasing the use of arrow functions, but they allow us to keep `this` in scope in the body of functions that are a part of a React class component

Here's a [quick article](https://medium.com/quick-code/react-quick-tip-use-class-properties-and-arrow-functions-to-avoid-binding-this-to-methods-29628aca2e25) that talks about it

react-app/src/components/Extract.js
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    onConfigFileChange(e) {
```
mgramigna, line 27:
```suggestion
    onConfigFileChange = (e) => {
```

Using an arrow function instead makes the `this` keyword a valid thing inside the scope of this function

This should be repeated for all of the functions in this component.

react-app/src/components/Extract.js
```diff
@@ -0,0 +1,106 @@
...
+        });
+    }
+
+    onSubmit(e) {
+        console.log(e);
+        console.log(this.configPath);
+    }
+
+    render() {
+        this.whatever = 10;
```
mgramigna, line 63: Assuming this variable doesn't get used for anything, can we remove it?

react-app/src/components/Extract.js
```diff
@@ -0,0 +1,106 @@
...
+                            <Form.Check type="checkbox" label="Filter by start date" />
+                        </Form.Group>
+                    </Col>
+                    <Col>
+                        <Form.Group controlId="formIncludeEndDate">
+                            <Form.Check type="checkbox" label="Filter by end date" />
+                        </Form.Group>
+                    </Col>
+                </Row>
+            </Form>
```
mgramigna, line 94: I noticed that in the form elements here, we never use `value` or `onChange` to sync up what is typed/selected into the form with the actual state variables. Since we have the state variables and the functions defined, maybe we should link them up via the `value` and `onChange` props of these components so the users' input is properly saved.

Note that the console log for configPath for instance always logs empty string since the value is never updated via `setState`.

react-app/src/components/Extract.js
```diff
@@ -0,0 +1,106 @@
...
+            <div className="nav-button-container">
+                <LinkContainer to="/">
+                    <button className="nav-button" variant="nav" id="nav-button">Cancel</button>
+                </LinkContainer>
+                <button className="nav-button" type="submit" variant="nav" id="nav-button" onClick={this.onSubmit}>Submit</button>
```
mgramigna, line 99: With my above comments to get rid of the CDNs and properly install/import bootstrap, you should now be able to use the bootstrap `Button` component and its variants properly.

I remember you mentioning that you intended to override some styles that bootstrap has on their buttons. We can talk about the details of that, but I still think we want to be consistent and use the bootstrap components where possible to avoid the need for more CSS.

react-app/src/components/Home.js
```diff
@@ -0,0 +1,26 @@
...
+import '../stylesheets/Home.css';
+
+class Home extends React.Component {
+    render() {
+        this.whatever = 10;
+        return <div>
+            <h1 className="page-title">Extractor UI</h1>
+            <div className="button-container">
+                <LinkContainer to="/extract">
+                    <button className="home-menu-button">Extract New</button>
```
mgramigna, line 18: Same comment as in `Extract.js` about using the bootstrap `Button` component.

react-app/src/components/Home.js
```diff
@@ -0,0 +1,26 @@
...
+
+import React from "react";
+import { LinkContainer } from 'react-router-bootstrap'
+// import { Button } from 'react-bootstrap'
+
+import '../stylesheets/Home.css';
+
+class Home extends React.Component {
+    render() {
+        this.whatever = 10;
```
mgramigna, line 13: Same comment about removing this unused variable

mgramigna (changes requested):
[MG1] These files could benefit from `lint` and `prettier` being run. For now, we can use the commands that we have provided in `package.json` to check both `lint` and `prettier` and fix any errors:

``` bash
npm run lint # checks for lint errors
npm run lint:fix # fixes any automatically fixable lint errors
npm run prettier # checks for any prettier formatting errors
npm run prettier:fix # automatically formats files
```

In the future, we can setup your editor to do some of this automatically, and also configure automated checking of these when a pull request is submitted, but that can be separate tasking

[MG2] When I pulled the code and ran the app, I got an error that `react-router-bootstrap` could not be found. It's possible you installed this locally, but without `--save` that would add it `react-app/package.json`. For any dependencies that you install (on the react side or electron side), we want to make sure they are listed in `dependencies` of the corresponding `package.json` so that when other people clone the code, their node_modules match up with yours. This will also produce changes in `package-lock.json`, which should be committed as well.

[MG others] Made a bunch of inline comments and suggestions that should be addressed. Let us know if you have any questions about them and we'd be happy to clarify.

react-app/src/components/Extract.js
SenecaAnderson: Fixed, I turned all the functions into arrow functions

react-app/public/index.html
SenecaAnderson: I have already installed bootstrap with npm install, does this mean I don't need this code at all?

react-app/src/components/Extract.js
SenecaAnderson: The main thing that I wanted to override was the color of the button. Bootstrap as already installed before, but I can't use any custom css with the variant option for the buttons because it gets overridden by built-in styles.

SenecaAnderson: I have replaced the <button> components with <Button>. I used the "danger" variant for the cancel button and the "info" variant for the submit button.

react-app/src/components/Extract.js
SenecaAnderson: This was a placeholder so I could start building the form without getting errors about not using "this" yet. I've deleted it.

react-app/src/components/Home.js
SenecaAnderson: Resolved by changing the Home page from a class to a function, since it doesn't need to use this or state.

package.json
```diff
@@ -51,7 +51,9 @@
     }
   },
   "dependencies": {
-    "electron-squirrel-startup": "^1.0.0"
+    "electron-squirrel-startup": "^1.0.0",
+    "react-router-bootstrap": "^0.25.0",
+    "react-router-dom": "^5.2.0"
```
mgramigna, line 56: React router bootstrap and react-router DOM only need to be in `react-app/package.json` they can be uninstalled from the root package.json

mgramigna (changes requested):
Just one last change request for the root `package.json`. Otherwise, looks good to me!

mgramigna:
For removing dependencies, it's best to use `npm uninstall --save <dependency-name>` this will

1) Remove it from `package.json` automatically
2) Update `package-lock.json` to re-generate without the newly uninstalled dependency.

Would you mind updating to do it this way? I'm assuming you just removed the packages from `package.json` manually. It's currently invalid JSON because of a trailing comma, and there are no `package-lock.json` changes.

mgramigna (approved):
Nice work!

jafeltra (commented):
This looks really great! I just had one additional question:

[JA1]: Is there a reason you have the various CSS files for `Home.css` and `Page.css`? I had thought some of the color choices and padding was something react-bootstrap would handle. If those classes just set what react-bootstrap sets, are the css files still needed? If you are including additional CSS beyond what react-bootstrap provides, that's fine too! I'm just curious about the reasoning.

SenecaAnderson:
The Page.css file has styles that are supposed to apply to many or every page in the application, while the Home.css file is styles that are specific to the home page. Most of the stuff that's there right now is button styling and positioning, because I wanted to change things like the color and size, and react-bootstrap has limited color options.

jafeltra (approved):
Sounds good! Thanks for the explanation! This looks great - and congrats on the first PR! 🎉

PR #2, opened 2021-06-03
Extractor API

SenecaAnderson:
Summary
Previously, the submit button on the Extract.js page did nothing. Now when the button is pressed, extraction runs successfully, so long as valid test data is provided. In order for it to run, two folders must exist: mcode-extraction-ui/test/config, which contains a config file called csv.config.json, and mcode=-extraction-ui/sample-client-data, which contains all the csv files and patient data. When extraction runs, a folder called 'output' is created. It contains the output files generated in extraction. The test and output folders are not tracked by git, so they're not included in this pull request.

New behavior
When you click the submit button on the Extract.js page, extraction runs, so long as valid test data is provided.

Code changes

There are changes to react-app/src/components/Extract.js, electron/preload.js, electron/electron.js, and electron/extraction.js. These changes mostly include function calls to each other.

Testing guidance

To test this, create a director called "test". Copy the config and sample-client-data folders from the test folder in mcode-extraction-framework, or retrieve test data from elsewhere; the important thing is that the config file is at test/config/csv.config.json, and that all necessary patient data is in a location pointed to by the config file. Run the program with:

npm start

Click "Extract New". Then click "Submit". The extraction process will run and output logs to the command line. Afterwards, there will be a new folder called "output" containing the result files.

### DIFF
--- a/electron/electron.js
+++ b/electron/electron.js
@@ -1,6 +1,7 @@
-const { app, BrowserWindow } = require('electron');
+const { app, BrowserWindow, ipcMain } = require('electron');
 const squirrel = require('electron-squirrel-startup');
 const path = require('path');
+const runExtraction = require('./extraction');
 
 // Handle creating/removing shortcuts on Windows when installing/uninstalling.
 if (squirrel) {
@@ -53,3 +54,11 @@ app.on('activate', () => {
 
 // In this file you can include the rest of your app's specific main process
 // code. You can also put them in separate files and import them here.
+
+ipcMain.handle('run-extraction', async (event, fromDate, toDate, configFilepath, runLogFilepath, debug, allEntries) => {
+  // run extractor - replace 'extractorCall()' with the actual call
+  // should be similar to node_modules/mcode-extraction-framework/src/cli/cli.js
+  console.log("inside extract handler");
+  const result = await runExtraction(fromDate, toDate, configFilepath, runLogFilepath, debug, allEntries);
+  return result;
+})
--- a/electron/electron.js
+++ b/electron/electron.js
@@ -56,9 +56,6 @@ app.on('activate', () => {
 // code. You can also put them in separate files and import them here.
 
 ipcMain.handle('run-extraction', async (event, fromDate, toDate, configFilepath, runLogFilepath, debug, allEntries) => {
-  // run extractor - replace 'extractorCall()' with the actual call
-  // should be similar to node_modules/mcode-extraction-framework/src/cli/cli.js
-  console.log("inside extract handler");
-  const result = await runExtraction(fromDate, toDate, configFilepath, runLogFilepath, debug, allEntries);
-  return result;
+  const extractedData = await runExtraction(fromDate, toDate, configFilepath, runLogFilepath, debug, allEntries).then((value) =>  value);
+  return extractedData;
 })
--- a/electron/electron.js
+++ b/electron/electron.js
@@ -56,6 +56,8 @@ app.on('activate', () => {
 // code. You can also put them in separate files and import them here.
 
 ipcMain.handle('run-extraction', async (event, fromDate, toDate, configFilepath, runLogFilepath, debug, allEntries) => {
-  const extractedData = await runExtraction(fromDate, toDate, configFilepath, runLogFilepath, debug, allEntries).then((value) =>  value);
+  const extractedData = await runExtraction(fromDate, toDate, configFilepath, runLogFilepath, debug, allEntries).then(
+    (value) => value,
+  );
   return extractedData;
-})
+});
--- a/electron/electron.js
+++ b/electron/electron.js
@@ -56,8 +56,6 @@ app.on('activate', () => {
 // code. You can also put them in separate files and import them here.
 
 ipcMain.handle('run-extraction', async (event, fromDate, toDate, configFilepath, runLogFilepath, debug, allEntries) => {
-  const extractedData = await runExtraction(fromDate, toDate, configFilepath, runLogFilepath, debug, allEntries).then(
-    (value) => value,
-  );
+  const extractedData = await runExtraction(fromDate, toDate, configFilepath, runLogFilepath, debug, allEntries);
   return extractedData;
 });
--- a/electron/extraction.js
+++ b/electron/extraction.js
@@ -4,31 +4,37 @@ const { MCODEClient } = require('../node_modules/mcode-extraction-framework/src/
 const logger = require('../node_modules/mcode-extraction-framework/src/helpers/logger');
 const { mcodeApp } = require('../node_modules/mcode-extraction-framework/src/cli/app');
 
-
 async function runExtraction(fromDate, toDate, configFilepath, runLogFilepath, debug, allEntries) {
-    try {
-        let defaultFromDate = fromDate;
-        let defaultToDate = toDate;
-        let defaultDebug = debug;
-        if (fromDate.length < 1) {
-            defaultFromDate = undefined;
-        }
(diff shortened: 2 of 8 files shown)


jafeltra:
I just merged in PR #1, which means this branch will need to be rebased so all the changes you made on that branch make it to this branch too! There are instructions for rebasing in the STEAM-branching document, but rebasing can be tricky, so we can also walk through it too!

SenecaAnderson:
This is rebased, so it's ready for review.

react-app/src/components/Extract.js
```diff
@@ -8,12 +9,18 @@ class Extract extends React.Component {
   constructor(props) {
     super(props);
 
+    const defaultPathToConfig = path.join('test', 'config', 'csv.config.json');
```
jafeltra, line 12: I made a comment about this above in the main comment box, but if we do rename the folder that has the config and sampel-client-data, we will need to update this path here.

jafeltra (commented):
This looks really good. It's super cool to see data being extracted. I had a couple comments inline and then one other question here:

[JA1] Since the `test` folder with the config file and sample data files isn't tracked by git, we should add the folder to the `.gitignore`. That will make it so the folder doesn't show up as an "untracked file" when you do git status and then we won't accidentally commit it in the future. Since it might be confusing to ignore a folder called "test" (often a test folder would contain unit tests and we might want to include some in the future), we might want to rename the folder to be `sample-extraction-data` or something similar, and then gitignore it.

electron/extraction.js
```diff
@@ -0,0 +1,40 @@
...
+      runLogFilepath,
+      defaultDebug,
+      allEntries,
+    ).then((value) => value);
+    return extractedData;
+  } catch (e) {
+    if (debug) logger.level = 'debug';
+    logger.error(e.message);
+    logger.debug(e.stack);
+    process.exit(1);
```
jafeltra, line 35: I just thought of one more thing. I'm not sure if we want to have the `process.exit(1)` here. If we do happen to have an error during extraction, the process.exit here will close the electron app completely. If we do hit an error, we probably want a way for the user to start the extraction process over again without needing to open the app again.

SenecaAnderson: We could have it redirect to an error page that tells the user what the problem was if that information is available, and gives them an option to go back to the extraction form. I could add that once the output display is done.

package.json
```diff
@@ -51,7 +51,10 @@
     }
   },
   "dependencies": {
-    "electron-squirrel-startup": "^1.0.0"
+    "bootstrap": "^5.0.1",
```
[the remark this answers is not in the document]
SenecaAnderson: I think I made this change (as well as removing react-bootstrap) for the other pull request, but it must have gotten lost during rebasing. I can remove the extra dependencies.

electron/extraction.js
jafeltra: Yeah I like that idea! I think it's fine to either include that page as part of the output task or as a separate task entirely. For this PR, I think removing this line and using the "Reset" button you already have should be sufficient!

package.json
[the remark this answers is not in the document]
jafeltra: Haha yeah that seems very possible. Thanks for updating!

electron/extraction.js
SenecaAnderson: I think that adding code for another button / display would take almost as much work as implementing the error page itself, but I couldn't implement the error page in this version because Extract is still a class component. Maybe I could just change the line to modify the value of this.state.submitted? That will achieve the same thing as hitting the "reset" button.

jafeltra: Oh I had just meant you could remove this line here and then the electron app wouldn't close. I think that would leave the app in the "running extraction..." state for now. And a user could use your existing "Reset" button to navigate back for now. Then we can implement the error page later on. Does that make sense?

electron/extraction.js
```diff
@@ -0,0 +1,40 @@
+// The code below is a copy of cli.js from mcode-extraction-framework
+
+const { MCODEClient } = require('../node_modules/mcode-extraction-framework/src/client/MCODEClient');
+const logger = require('../node_modules/mcode-extraction-framework/src/helpers/logger');
+const { mcodeApp } = require('../node_modules/mcode-extraction-framework/src/cli/app');
```
jafeltra, line 5: Because you installed these using NPM and because of how we set up the exports from the MEF, you don't need the file path to node_modules here. Instead, you can just import directly from `mcode-extraction-framework` like:

```
const { logger, mcodeApp, MCODEClient } = require('mcode-extraction-framework');
```

There are examples of this in the Base ICARE Client repository [here](https://github.com/mcode/base-icare-extraction-client/blob/main/src/app.js#L5-L8) in case you are curious to see it done in other places!

electron/electron.js
```diff
@@ -53,3 +54,10 @@ app.on('activate', () => {
 
 // In this file you can include the rest of your app's specific main process
 // code. You can also put them in separate files and import them here.
+
+ipcMain.handle('run-extraction', async (event, fromDate, toDate, configFilepath, runLogFilepath, debug, allEntries) => {
+  const extractedData = await runExtraction(fromDate, toDate, configFilepath, runLogFilepath, debug, allEntries).then(
+    (value) => value,
```
jafeltra, line 60: I think that the `.then` statement here isn't necessary. Because you're using the `await` keyword when calling `runExtraction` this will wait until the promise is resolved in `runExtraction` before the `extractedData` variable is assigned. This makes it so there isn't a promise that needs to wait for a `.then` here, so you should be safe to remove it.

jafeltra (commented):
I had other comments. Sorry for the multiple groups they've come in.

electron/extraction.js
```diff
@@ -0,0 +1,39 @@
+// The code below is a copy of cli.js from mcode-extraction-framework
+
+const { MCODEClient } = require('mcode-extraction-framework/src/client/MCODEClient');
+const logger = require('mcode-extraction-framework/src/helpers/logger');
+const { mcodeApp } = require('mcode-extraction-framework/src/cli/app');
```
jafeltra, line 5: I think you can use `require('mcode-extraction-framework')` for each of these, rather than also include the file path to the MCODEClient, logger, and app. The way we set up the exports from the MEF should let you import each directly. That would let you also combine this to one line, like this:

```
const { logger, mcodeApp, MCODEClient } = require('mcode-extraction-framework');
```

jafeltra (commented):
This is looking really good! I just left one more comment about the imports from MEF. If that approach doesn't work for some reason, let me know!

jafeltra (approved):
This looks great! It's so cool to see the extracted data (in the console).

julianxcarter (changes requested):
This looks great to me! I have two comments that really just minor edits, but overall everything worked like a charm.

electron/extraction.js
```diff
@@ -0,0 +1,35 @@
...
+      defaultDebug,
+      allEntries,
+    ).then((value) => value);
+    return extractedData;
+  } catch (e) {
+    if (debug) logger.level = 'debug';
+    logger.error(e.message);
+    logger.debug(e.stack);
+  }
+  return undefined;
```
Dtphelan1, line 32: DP1 - we could probably move this into the catch block, since the only time we'll get to this line is when the catch statement has triggered.

Dtphelan1 (commented):
This looks good to me beyond that last comment. Once that's changed, I'm good to approve.

julianxcarter (approved):
Looks great, awesome work! @SenecaAnderson

PR #3, opened 2021-06-15
Output display v2

SenecaAnderson:
### Summary
After extraction runs, the results display in a Results page. The user is able to toggle between the JSON schemas for each patient. Additionally, the theme colors are different.

### New Behavior
After extraction completes, the app redirects to a page where the user can view the results. Additionally, the theme colors are different.

### Code Changes

- Sass was installed in order to change built-in bootstrap variants like primary, secondary, info, etc. and to define theme color variables. 
- The CSS files for various pages were modified, deleted, and/or added. 
- react-bootstrap was updated from version 1.6.1 to 2.0.0.beta-0. 
- ResultPage, ResultSidebar, ResultHeader, and PatientData were all modified in order to create the output display. 
- The styling of buttons and titles on various pages was also modified to fit the new theme colors.

### Testing Guidance
Start the program with npm start. Click the "Extract New" button. Either modify the config file pathway to point to a valid configuration file, or create a config file at "code-extraction-ui/sample-extraction-data/csv.config.json" and use the default value. Click "Submit" to run extraction. After a moment during which a processing page may appear, the app will redirect to the Results page. Click on the sidebar dropdowns ("Patient 0", "Patient 1", etc.) in order to toggle between the JSON schemas. Click "Exit" to return the the Extract New page. Clicking "Save" does nothing.


react-app/src/App.js
```diff
@@ -1,23 +1,27 @@
...
 import { BrowserRouter as Router, Switch, Route } from 'react-router-dom';
 
 import Home from './components/Home';
 import Extract from './components/Extract';
+import ResultPage from './components/ResultPage';
 
-import 'bootstrap/dist/css/bootstrap.min.css';
-import './App.css';
-import './stylesheets/Page.css';
+// import 'bootstrap/dist/css/bootstrap.min.css';
```
mgramigna, line 8: Assuming this line is unneeded now since you're using the boostrap sass
```suggestion
```

react-app/src/App.js
```diff
@@ -1,23 +1,27 @@
...
+import './stylesheets/custom.scss';
+import './stylesheets/Page.scss';
 
 function App() {
+  const [extractedData, setExtractedData] = useState(['test', 'test', 'test']);
```
mgramigna, line 13:
```suggestion
  const [extractedData, setExtractedData] = useState([]);
```

Can we have the default here just be empty array? It's later populated via actual extraction so it shouldn't hurt anything

react-app/src/components/PatientData.js
```diff
@@ -0,0 +1,21 @@
...
+        <div className="page-text">
+          <p>Select a patient to view their information.</p>
+        </div>
+      )}
+      {props.id >= 0 && (
+        <div className="patient-json-display">
+          <ReactJson src={props.patientJson} collapsed={3} />
+        </div>
+      )}
+    </div>
```
mgramigna, line 17: I'm assuming the `< 0` is used here to basically check if we have nothing yet since you initialize the state variable to -1 in another file.

I think we should be careful in assuming that ID is a number. For example, if `props.id` is "matt-123", it may still pass this boolean condition, but things get a little dicey.

I'll have a companion comment in the file where the `id` prop is actually initialized and passed along, but basically I think we should be useing `=== null` and `!== null` as our cases here instead of comparing against 0. And then in the other component where the patient ID state variable is initialized we can default it to null (see other comment)

react-app/src/components/ResultPage.js
```diff
@@ -0,0 +1,28 @@
+import React, { useState } from 'react';
+import { Col, Container, Row } from 'react-bootstrap';
+// import ResultHeader from './ResultHeader';
+import PatientData from './PatientData';
+import ResultSidebar from './ResultSidebar';
+
+import '../stylesheets/ResultPage.scss';
+
+function ResultPage(props) {
+  const [patientID, setPatientID] = useState(-1);
```
mgramigna, line 10: Per comment in `PatientData.js`, let's initialize this state variable to null instead of -1. That makes things a little cleaner.

```suggestion
  const [patientID, setPatientID] = useState(null);
```

react-app/src/components/ResultSidebar.js
```diff
@@ -0,0 +1,43 @@
...
+        <Button className="nav-button" siz="lg" variant="outline-secondary" onClick={onSave}>
+          Save
+        </Button>
+      </div>
+    </div>
+  );
+}
+
+// DELETE ME
+//       <h1 className="page-subtitle">Results</h1>
```
mgramigna, line 42: As you wish

```suggestion
```

mgramigna (changes requested):
Overall this looks awesome. I just have a few small change requests, nothing major. Nice work

mgramigna (approved):

react-app/src/stylesheets/Page.scss
```diff
@@ -0,0 +1,47 @@
+@import './_variables.scss';
+/* .page {
+    background-color: white;
+    background-size: cover;
+    height: 100%;
+} */
```
Dtphelan1, line 6:
```suggestion
```

react-app/src/stylesheets/Page.scss
```diff
@@ -0,0 +1,47 @@
...
+  flex-wrap: nowrap;
+}
+.col-xl-3,
+.col-lg-3,
+.col-md-3,
+.col-sm-3,
+.col-3 {
+  padding-left: 0px;
+  padding-right: 0px;
+}
```
Dtphelan1, line 13: DP1 -  I think there is a `no-gutters` className in bootstrap that we could be using here instead of overriding the .col-*-3 padding.

react-app/src/stylesheets/custom.scss
```diff
@@ -0,0 +1,12 @@
...
+$theme-colors: (
+  'primary': $primary,
+  'secondary': $secondary,
+  'info': $info,
+  'success': $success,
+  'warning': $warning,
+  'danger': $danger,
+);
+
+@import '../../node_modules/bootstrap/scss/bootstrap.scss';
```
Dtphelan1, line 12: DP2 - We shouldn't be relative pathing to node_modules in any of our JS/SCSS files. SCSS used through CRA supports pathing to the `node_modules` directory using the `~` character. See [here](https://create-react-app.dev/docs/adding-a-sass-stylesheet/)

Dtphelan1 (changes requested):

react-app/src/stylesheets/Page.scss
Dtphelan1: This is a pretty nifty cheatsheet that I often use when in react-bootstrap land: https://hackerthemes.com/bootstrap-cheatsheet/#no-gutters

react-app/src/components/Result.js
```diff
@@ -0,0 +1,22 @@
...
+      </Accordion.Header>
+    </Accordion.Item>
+  );
+}
+
+export default Result;
+
+// <ListGroup.Item action onClick={onSelectPatient} eventKey={props.id} id={props.id} variant="secondary">
+//  Patient {props.id}
+// </ListGroup.Item>;
```
Dtphelan1, line 22:
```suggestion
```

Dtphelan1 (approved):

PR #4, opened 2021-06-17
Logger

SenecaAnderson:
### Summary
A custom transport was added to the logger in order to retrieve the logged messages. The logged messages are passed back to the renderer process and stored in App.js.

### New Behavior
Logged messages are retrieved and sent back to the renderer process.

### Code Changes

- An array to hold the logged messages is created in electron.js
- A custom transport class was created, and an instance of the transport is added to the logger in electron.js
- electron.js returns an object containing both extractedData and loggedMessages
- The useSubmit function in Extract.js saves the loggedMessages using a function passed from App.js as a prop
- The dependency strip-ansi was added to format the messages
- When an error occurs, extraction.js returns null instead of undefined

### Testing Guidance
Start the app with npm start. Run extraction using a valid config file and data. The object, extractedData, and loggedMessages will be outputted to the console when extraction completes.

### DIFF
--- a/electron/Transport.js
+++ b/electron/Transport.js
@@ -1,0 +1,14 @@
+const Transport = require('winston-transport');
+
+module.exports = class CustomTransport extends Transport {
+  constructor(opts) {
+    super(opts);
+    this.loggedMessages = opts.loggedMessages;
+  }
+
+  log(info, callback) {
+    this.loggedMessages.push(`${info.timestamp} [${info.level}]: ${info.message}`);
+    // Perform the writing to the remote service
+    callback();
+  }
+};
--- a/electron/Transport.js
+++ b/electron/Transport.js
@@ -1,4 +1,5 @@
 const Transport = require('winston-transport');
+const stripAnsi = require('strip-ansi');
 
 module.exports = class CustomTransport extends Transport {
   constructor(opts) {
@@ -7,7 +8,8 @@ module.exports = class CustomTransport extends Transport {
   }
 
   log(info, callback) {
-    this.loggedMessages.push(`${info.timestamp} [${info.level}]: ${info.message}`);
+    const level = stripAnsi(info.level);
+    this.loggedMessages.push(`${info.timestamp} [${level}]: ${info.message}`);
     // Perform the writing to the remote service
     callback();
   }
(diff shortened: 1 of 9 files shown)


electron/Transport.js
```diff
@@ -0,0 +1,16 @@
...
+    this.loggedMessages = opts.loggedMessages;
+  }
+
+  log(info, callback) {
+    const level = stripAnsi(info.level);
+    this.loggedMessages.push(`${info.timestamp} [${level}]: ${info.message}`);
+    // Perform the writing to the remote service
+    callback();
+  }
+};
```
Dtphelan1, line 16: DP1 - Could we rename the CustomTransport class to be a little more informative? Something like InMemoryTransport – whatever you think gets across the spirit of what this transport is accomplishing. 

DP2 - Could you also rename the file to be the name of the class we define, whatever that ends up being in resolving DP1. This will require updating imports as well.

electron/Transport.js
```diff
@@ -0,0 +1,16 @@
...
+
+module.exports = class CustomTransport extends Transport {
+  constructor(opts) {
+    super(opts);
+    this.loggedMessages = opts.loggedMessages;
+  }
+
+  log(info, callback) {
+    const level = stripAnsi(info.level);
+    this.loggedMessages.push(`${info.timestamp} [${level}]: ${info.message}`);
```
Dtphelan1, line 12: DP3 - Now that we have the stripAnsi function, do we need to handmake this message anymore? Is there some way of getting this fully formatted log message off of the `info` object, instead of recreating the format ourselves?

react-app/src/components/Extract.js
```diff
@@ -53,7 +53,11 @@ function Extract(props) {
     // allEntries parameter is true if it's not filtered by date, false if it is
     const filter = !(filterStart || filterEnd);
     window.api.extract(fromDate, toDate, configPath, logPath, includeDebug, filter).then((value) => {
-      props.setExtractedData(value);
+      console.log(value);
+      console.log(value.extractedData);
+      console.log(value.loggedMessages);
```
Dtphelan1, line 58: DP4 - Let's drop these console logs before putting the code onto master.

Dtphelan1 (changes requested):

Dtphelan1 (approved):
Great work!

julianxcarter (approved):
This looks great!

PR #5, opened 2021-06-18
Logger display

SenecaAnderson:
### Summary
The result page now has an option to display the logger messages from extraction.

### New Behavior
When the result page appears, the first item in the sidebar is labelled "Log File". When the user clicks on it, a list of logged messages displays on the right.

### Code Changes
Changes were made to ResultPage.js, ResultSidebar.js, and PatientData.js in order to display the log file option in the sidebar, toggle to it, and display the messages when it's selected.

### Testing Guidance
Run the app with npm start. Click "Extract New" and run extraction with valid data. In the result page, click "Log File".


package-lock.json
```diff
@@ -29,6 +29,35 @@
...
+        "cross-env": "^7.0.0",
+        "eslint": "^6.6.0",
+        "eslint-config-airbnb-base": "^14.0.0",
+        "eslint-plugin-import": "^2.18.2",
+        "jest": "^26.6.3",
+        "jest-when": "^2.7.0",
+        "nock": "^11.7.0",
+        "rewire": "^5.0.0"
+      }
+    },
```
jafeltra, line 60: This relative path feels odd to have in a package-lock.json file. It doesn't look you have any relative paths in the `package.json` so I don't know why this got added in the branch. Do you think you can try re-running an `npm install` and see if this gets removed?

react-app/src/components/ResultPage.js
```diff
@@ -14,10 +14,18 @@ function ResultPage(props) {
       <Container fluid>
         <Row>
           <Col xxl={3} xl={3} lg={3} md={3} sm={3} xs={3} className="sidebar-col">
-            <ResultSidebar extractedData={props.extractedData} setPatientID={setPatientID} />
+            <ResultSidebar
+              extractedData={props.extractedData}
+              setPatientID={setPatientID}
+              loggedMessages={props.loggedMessages}
```
jafeltra, line 20: It doesn't look like `ResultsSidebar` is using `props.loggedMessages` at all. Is there any reason to keep it passed into the component? Or can we remove it?

react-app/src/components/ResultSidebar.js
```diff
@@ -23,6 +23,15 @@ function ResultSidebar(props) {
...
         <Accordion defaultActiveKey="0" flush>
+          <Accordion.Item eventKey={-1}>
+            <Accordion.Header
+              onClick={() => {
+                props.setPatientID(-1);
+              }}
+            >
+              Log File
+            </Accordion.Header>
+          </Accordion.Item>
```
jafeltra, line 34: This is more of a general UI interaction question, but is there anything that will be expanded in the Log File item in this accordion? I know we want to include the resource count for each patient, which will be displayed when the accordion item expands for those items. If there isn't anything similar we want to expand for the log file, does it make sense to move this out of the accordion and have it just be a button/label/some other clickable item above the accordion for the patient results?

react-app/src/components/PatientData.js
```diff
@@ -1,15 +1,22 @@
...
         <div className="page-text">
-          <p>Select a patient to view their information.</p>
+          <p>Select a patient to view their information or click on "Log File" to view the logger messages.</p>
         </div>
       )}
-      {props.id !== null && (
+      {props.id !== null && props.id === -1 && (
+        <div>
+          <h3 className="page-subtitle">Log File</h3>
+          <LogList loggedMessages={props.loggedMessages} />
```
jafeltra, line 16: This is pretty minor, but is it possible to have the list of log messages scroll while keeping the "Log File" title static at the top of the screen? I think that might be nice to have since the list is usually so long and the "Log File" feels like a header that usually stays static at the top of the page. If that's really difficult CSS-wise, we can evaluate if it's actually necessary.

jafeltra (commented):
This looks good! I had a few comments inline and one other general question. Let me know if you want to talk about any of them!

[JA1] - This doesn't seem to have been introduced in this PR, but is there a reason we have a blank `Logs.js` file? Are we planning to use that in the future for anything? It doesn't look like it was introduced in this PR so maybe there was discussion that I missed about it.

package-lock.json
SenecaAnderson: Running npm install didn't change it. I think this might be from when I was testing the code I changed to fix the debug error, since I changed the path of the framework in package.json to point to the local copy of the framework. I changed the package.json file back after I finished testing, so I'm not sure why the package-lock.json file hasn't updated. Since it's auto-generated, can I just delete package-lock.json and run npm install again?

react-app/src/components/ResultSidebar.js
SenecaAnderson: I put it in the accordion because I wanted it to have the same styling as the rest of the items in the list, and because otherwise, it creates some minor issues with toggling between the log file and patient data. When the log file isn't an Accordion.Item, clicking on it doesn't deactivate the patient data that was selected before, even though it changes the display, so the name of the patient remains highlighted. Additionally, the areas of the ListGroup.Item that are clickable are different from from the areas of the Accordion.Item that are clickable, which may annoy users. 

I have a way to remove the arrow with CSS, so I could do that instead of making it a separate component. That way it can keep all the functionality of the accordion without looking like a dropdown.

react-app/src/components/PatientData.js
SenecaAnderson: I tried making the title sticky, but I think the grid system interferes with the display somehow. The ListGroup covers it up so that the sticky title only appears while scrolling.

react-app/src/components/ResultSidebar.js
SenecaAnderson: It turns out that the method for removing the arrow icon removes the arrow icon from every accordion item, because all the arrows are have the CSS style .accordion-button::after. Do you think it's better to keep the arrow, or to make the log file a ListGroup.Item?

package-lock.json
jafeltra: You're right that it is autogenerated, so you could delete it and it would regenerate it. I think before you try that approach, I would remove the `node_modules` folder and try running `npm install` again. Sometimes if the local link is still linked in your `node_modules`, `package-lock.json` doesn't update correctly. I don't know the real reason for this, but I think it has something to do with the local linking. So try deleting `node_modules` from the root directory and then run `npm install` and it will hopefully update `package-lock.json`.

react-app/src/components/ResultSidebar.js
jafeltra: After talking at stand up, we'll leave the log file in the accordion, and I've created a separate Jira task [here](https://jira.mitre.org/browse/STEAM-631) for including the log statistic summary that we will implement later (after prioritizing it).

package-lock.json
SenecaAnderson: I tried deleting node_modules and running npm install, then deleted package-lock.json and ran npm install. That fixed the problem.

jafeltra: So deleting `node_modules` and running `npm install` alone didn't fix the problem?

SenecaAnderson: No, it still had the relative path. I think that part of the package-lock.json file didn't get updated for some reason.

react-app/src/components/PatientData.js
jafeltra: Hm I feel like there should be a way to make the header stay at the top of the screen. I think having it stay at the top will make it feel a little nicer for users too. If the grid system doesn't let the top "Log File" header stay sticky, there is probably a way to set the height of the overall page so that the page stays at a 100% height and doesn't scroll, and then you can set the div that contains the Log messages (or the top level div within the LogList component) so that that div takes up the remaining space on the screen and any overflow within that should scroll.

If you want any help trying out those approaches, let me know.

package-lock.json
jafeltra: Hm that's odd. When you deleted the `package-lock.json` and regenerated it with `npm install` are there a lot of changes in it? Or does it just remove that one section?

SenecaAnderson: It looks like the "resolved" and "integrity" lines of each dependency were deleted, and "license" : "MIT" was added. Additionally, I just did a search in the package.json while comparing the old and new files, and it looks like the name of the line "mcode-extraction-framework": "git+https://github.com/mcode/mcode-extraction-framework.git#develop", was somehow replaced with "generic-button": "git+https://github.com/mcode/generic-button.git#develop". I renamed the nav-button CSS class a while ago using find and replace, and I think that I accidentally clicked something while doing that and caused this error.

SenecaAnderson: The current version of package.json has the correct dependency.

react-app/src/components/PatientData.js
SenecaAnderson: I tried to make the title sticky in the ExtractionError component of the error-page branch, which doesn't use the grid system, and it looks like the issue is actually that the ListGroup covers the sticky component. Additionally, I realized while looking into this that the sticky title is basically a NavBar with a white background. Maybe I could use a NavBar instead? Since it's a react-bootstrap component, maybe the ListGroup issue will go away.

package-lock.json
jafeltra: Hmmm that is really odd. So nowhere in your `package-lock.json` has the usual "mcode-extraction-framework": "git+https://github.com/mcode/mcode-extraction-framework.git#develop" line? Because if the package-lock.json file was regenerated, it's pretty likely that a lot of lines would change, so it doesn't seems surprising that diff is having trouble showing the change, but it should be _somewhere_. And I'm not sure what that generic-button is for.

react-app/src/components/PatientData.js
jafeltra: Making a NavBar with a white background is definitely another approach! I think typically NavBars are used for a full app, rather than just a title on a specific page, so I'm not completely sure if that's the approach we want to go with in the end, but it's definitely worth exploring and seeing generally other ways to position the elements.

The thought process I was having earlier was to make sure that each page of the whole app only takes up 100% of the height of the app. Then any components on that page that need more room, like the LogList, can scroll within that space, so they don't extend the page any longer. Just another possible approach to explore!

package-lock.json
SenecaAnderson: While styling the form in the file-and-date-pickers branch, I changed the name of the nav-button CSS class to generic-button, because I used react-bootstrap Buttons for the file pickers, and I wanted the same styling, so the name "nav-button" no longer made sense. I used find-and-replace, so I think that I somehow accidentally replaced text in package.json after I switched to the logger-display branch. Both package.json and package-lock.json have the correct code-extraction-framework line now.

react-app/src/components/PatientData.js
SenecaAnderson: Alright, that makes sense. I'll try implementing that.

package-lock.json
SenecaAnderson: It looks like I misread the package-lock.json file while comparing; I was on the search tab while looking at the two versions, not the source control tab, so it showed the generic-button replacement, and I thought it was the old version. I looked at the file again and realized that while the correct file path was in one spot, the local path was at another, so I deleted node_modules and package-lock.json and ran npm install while both were deleted. This fixed the problem.

jafeltra: The other very simple approach to this problem could just be doing a `git reset origin/master package-lock.json` to reset the file, then you could remove `node_modules` and run `npm install` again to make sure nothing needed to change, which it shouldn't have since this branch didn't change any dependencies. That might be cleanest solution to this.

react-app/src/components/PatientData.js
SenecaAnderson: Solved this by moving the subtitle after the ListGroup so that the ListGroup scrolls behind it instead of in front of it.

package-lock.json
SenecaAnderson: Is this something I should do in addition to what I already did, or just an alternative if I run into this problem again in the future?

jafeltra: That's just something you can do in the future if you end up with package-lock.json changes that you know you don't want to keep. Usually, you should not have any package-lock.json changes if you haven't changed package.json. Ideally, when you don't change package.json, you can `npm install` or `npm ci` (which is installing based on package-lock directly) and the install will use what is listed in the package-lock so it grabs the correct versions of packages.

In this case, I think it's okay to have upgraded the packages you have listed here.

react-app/src/components/PatientData.js
jafeltra: Oh interesting. I pulled the branch and this definitely works how I was expecting it to in the app. However, I think having the `<h3>` below the `<LogList>` is confusing to read in the code. Is there a way to make just the <LogList> component scrollable so we can maintain the order in the code that matches the rendered order in the app?

react-app/src/components/ResultPage.js
jafeltra: Just as a reminder, I think you mentioned you were going to address this comment and the JA1 comment, but I haven't seen those changes get pushed yet. No rush, just wanted to make sure they were still on your radar!

SenecaAnderson: Looks like I stashed those changes instead of committing them. I'll push them now.

react-app/src/components/ResultPage.js
```diff
@@ -14,18 +14,10 @@ function ResultPage(props) {
...
-              setPatientID={setPatientID}
-              loggedMessages={props.loggedMessages}
-            />
+            <ResultSidebar extractedData={props.extractedData} setPatientID={setPatientID} />
           </Col>
           <Col>
-            <PatientData
-              patientJson={props.extractedData[patientID]}
-              id={patientID}
-              loggedMessages={props.loggedMessages}
```
jafeltra: `PatientData` does use the `loggedMessages` prop, so we need to still pass that one onto this component. Right now, without it, the app crashes when I try to view the Log File.

react-app/src/components/PatientData.js
jafeltra: Cool! This looks much nicer to me, and the code is much more readable! Thanks for spending the time to look into this!

react-app/src/components/PatientData.js
```diff
@@ -1,15 +1,22 @@
...
         </div>
       )}
-      {props.id !== null && (
+      {props.id !== null && props.id === -1 && (
+        <div>
+          <h3 className="page-subtitle sticky-item">Log File</h3>
+          <LogList loggedMessages={props.loggedMessages} />
+        </div>
+      )}
+      {props.id !== null && props.id >= 0 && (
```
mgramigna, line 19: Per my comments on the previous PR, I think we should try to avoid using -1 here for the patient ID.

Really what the -1 is representing here is a different piece of information, which is that we wish to view the Logs rather than any patient data. I think this should be controlled in a separate state variable, something like `[showLogs, setShowLogs]` which is a boolean that indicates whether we are viewing the logs instead of patient data. Let me know if you have questions about this. There is one other comment I'll make as a companion to this one

react-app/src/components/ResultSidebar.js
```diff
@@ -23,6 +23,15 @@ function ResultSidebar(props) {
       <ResultHeader />
       <div className="sidebar-interior">
         <Accordion defaultActiveKey="0" flush>
+          <Accordion.Item eventKey={-1}>
+            <Accordion.Header
+              onClick={() => {
+                props.setPatientID(-1);
```
mgramigna, line 29: Per my other comment about the `[showLogs, setShowLogs]` variable, we should be setting patientID to `null` here, and then set the showLogs variable to `true` since that's what we are doing.

jafeltra (approved):
This looks great! Thanks for making the changes. This looks really nice!

mgramigna (changes requested):
Looks good. 2 small comments

react-app/src/components/PatientData.js
```diff
@@ -1,15 +1,22 @@
...
         </div>
       )}
-      {props.id !== null && (
+      {props.showLogs && (
+        <div>
+          <h3 className="page-subtitle sticky-item">Log File</h3>
+          <LogList loggedMessages={props.loggedMessages} />
+        </div>
+      )}
+      {props.id !== null && props.id >= 0 && (
```
mgramigna, line 19: I think this last props.id >= 0 is unnecessary now / could be replaced with `!props.showLogs`

jafeltra (approved):
This looks good to me! Nice work!

mgramigna (approved):

PR #6, opened 2021-06-21
Error page

SenecaAnderson:
### Summary
When extraction fails, an error page will display. It explains that there was an error, displays the logged messages, and provides a button back to the Extract component.

### New Behavior
If the config file name or contents are invalid, or extraction fails for some reason, the error page will appear.

### Code Changes
A new ExtractionError component was added. Additionally, there are some minor changes to the page-text css class.

### Testing Guidance
Start the app with npm start. Click on "Extract New". There are two options to deliberately cause errors:

- Add a typo to the default config file path, or provide any invalid file path. There will be only one logged message if you do this.
- Modify the config file so that something will cause an error. If you change the last extractor file path, this will cause a long list of logged messages to appear, so you will get to see what the error page looks like with many messages.

Click "Submit". After the error occurs, the error page will appear, and you can view the logged messages. Click on "Extract New" to run the process again. If you would like to see debug-level messages, check "Log output debugging information". Note that once you set the debug flag, the flag will be set until you restart the app from the terminal, so every extraction after that will have debug-level messages.


jafeltra:
Can you update the base branch of this PR to be `logger-display`? Since this branch has changes on top of that branch, it will show just the relevant differences in this PR so the right changes are reviewed ehre, and when that branch gets merged to master, the base branch of this PR will update. That will also prevent us from accidentally merging this to `master` before the `logger-display` PR is merged.

electron/Transport.js
```diff
@@ -0,0 +1,14 @@
...
+    super(opts);
+    this.loggedMessages = opts.loggedMessages;
+  }
+
+  log(info, callback) {
+    this.loggedMessages.push(`${info.timestamp} [${info.level}]: ${info.message}`);
+    // Perform the writing to the remote service
+    callback();
+  }
+};
```
jafeltra, line 14: It looks like maybe an old version of the `InMemoryTransport` got added back. Is this intentional? Or can we remove this file?

react-app/src/components/ExtractionError.js
```diff
@@ -0,0 +1,26 @@
...
+        checked.
+      </p>
+      <div className="button-container">
+        <LinkContainer to="/extract">
+          <Button className="nav-button" size="lg" variant="outline-secondary">
+            Extract New
+          </Button>
+        </LinkContainer>
+      </div>
+      <LogList loggedMessages={props.loggedMessages} />
```
jafeltra, line 21: Similar to my comment on the `LogList`, but I think it might be nice of the heading and directions you have here stayed at the top of the page and just the list of log messages scrolls. It might be possible to make that change within the `LogList` component, so maybe it would just come for free if you made the change in the `logger-display` branch and updated this branch. If so, then this should be set! Otherwise, it might be nice to add a similar change to this page as well.

jafeltra (changes requested):
I noticed a couple small things, but overall this is looking really good! It's cool to see so many pages implemented now!

electron/Transport.js
SenecaAnderson: The old version must have been added back in during rebasing. I've deleted it.

react-app/src/components/ExtractionError.js
SenecaAnderson: I tried making the subtitle sticky. However, I ran into the same issue as on the logger-display branch--the LogList item covers it up when you scroll.

jafeltra: It sounds like you're doing some exploration for this on the `logger-display` branch. Let's see if we can get it working as expected over on that branch, and then we can evaluate what we'll need to update here!

SenecaAnderson: I created a CSS class called sticky-item in the logger-display branch. Once that branch is merged and this one is rebased, I can apply that styling to the subtitle of the error page as well.

jafeltra: Since this branch is based off the `logger-display` branch, I actually think  you could rebase this branch against the updated version, and you should just get the one or two new commits from that branch on this one. If that rebase doesn't go smoothly though, we can definitely just hold off until the `logger-display` branch is merged like you said!

mgramigna:
This looks good to me once the rebase goes through

react-app/src/components/PatientData.js
```diff
@@ -5,10 +5,10 @@ import LogList from './LogList';
 function PatientData(props) {
   return (
     <div>
-      {props.id === null && !props.showLogs && (
```
jafeltra, line 8: I think we still want to have the `!props.showLogs` condition in this statement, right? I'm guessing it just got lost in a rebase.

react-app/src/components/ExtractionError.js
jafeltra: I think it would be good to carry over the scrolling in the LogList component from the `logger-display` branch to here. When I have the app window small and a few different errors listed, the title and directions scroll up on the page when I scroll down to look at the rest of the errors. I think it would be nice to have the same interaction that the LogFile view has.

jafeltra (commented):
This is looking good, I just had a few other small comments inline and one that I couldn't add to a specific line:

[JA1] - I'm not sure if some CSS property got lost in the rebasing, but the "Log File" title on the Results page isn't aligned on the left side like it is on the `logger-display` branch.

jafeltra (approved):
This looks good to me! Thanks for looking into that scroll, and I think it is okay as it is now since there likely won't be many errors listed on that page since it usually crashes pretty early if there are fatal errors!

mgramigna (approved):

PR #7, opened 2021-06-29
File and date pickers

SenecaAnderson:
### Summary
The form on the Extract page was modified. The config file path and log file path fields now have file selectors instead of text inputs. When the user checks "Filter by start date" or "Filter by end date", a date picker appears under the checkbox, allowing the user to chose to and from dates.

### Code Changes
The form on the Extract page was modified to use file and date pickers, and to update the toDate and fromDate variables based on input. preload.js and electron.js were modified to add functions to handle file selection.

### New Behavior
The user selects config and log files using a file picker. The user is able to filter results by date.

### Testing Guidance
Start the app with npm start. Click on "Extract New". Click on "Select Config File" or "Select Log File" to open the file selectors and choose files. Check "Filter by start date" to make the date picker for fromDate appear. Check "Filter by end date" to make the date picker for fromDate appear. Any combination of values (dates, log file, etc) that would work for the command line interface should also work in the UI.

### DIFF
--- a/electron/electron.js
+++ b/electron/electron.js
@@ -1,4 +1,5 @@
 const { app, BrowserWindow, ipcMain } = require('electron');
+const { dialog } = require('electron');
 const squirrel = require('electron-squirrel-startup');
 const path = require('path');
 const { logger } = require('mcode-extraction-framework');
@@ -13,12 +14,15 @@ if (squirrel) {
   app.quit();
 }
 
+// declare mainWindow ahead of time so that it can be accessed in .handle('get-file')
+let mainWindow;
 const createWindow = () => {
   // Create the browser window.
-  const mainWindow = new BrowserWindow({
+  mainWindow = new BrowserWindow({
     width: 800,
     height: 600,
     webPreferences: {
+      enableRemoteModule: true,
       preload: path.join(__dirname, './preload.js'),
     },
   });
@@ -65,3 +69,4 @@ ipcMain.handle('run-extraction', async (event, fromDate, toDate, configFilepath,
   const extractedData = await runExtraction(fromDate, toDate, configFilepath, runLogFilepath, debug, allEntries);
   return { extractedData, loggedMessages };
 });
+ipcMain.handle('get-file', async () => dialog.showOpenDialog(mainWindow, { properties: ['openFile'] }));
--- a/electron/electron.js
+++ b/electron/electron.js
@@ -22,7 +22,6 @@ const createWindow = () => {
     width: 800,
     height: 600,
     webPreferences: {
-      enableRemoteModule: true,
       preload: path.join(__dirname, './preload.js'),
     },
   });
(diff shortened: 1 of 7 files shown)


electron/electron.js
```diff
@@ -13,12 +14,15 @@ if (squirrel) {
...
+// declare mainWindow ahead of time so that it can be accessed in .handle('get-file')
+let mainWindow;
 const createWindow = () => {
   // Create the browser window.
-  const mainWindow = new BrowserWindow({
+  mainWindow = new BrowserWindow({
     width: 800,
     height: 600,
     webPreferences: {
+      enableRemoteModule: true,
```
mgramigna, line 25: Is there a reason this was added? I believe that the api/preload style that we've set up avoids the need for remote modules. Notably, this also introduces some security concerns into the app if not handled properly.

I think this can be removed, the app should still work.

SenecaAnderson: I think I added this while trying to resolve a bug with the dialog, then forgot to change it back later when I found the solution. I can delete it.

mgramigna (changes requested):
* I think it would be nice to have a way to remove a selected file (via some type of delete button). That way if I upload say a previous logfile but say I don't want that anymore, I can delete it without having to refresh the app. 
* I think upload buttons should look a little different. My suggestion would be to have just a rendering of the filename with actual upload/delete buttons as needed. Here's an example of what that might look like (cc'ing @Dtphelan1): 
image

mgramigna:
> * I think it would be nice to have a way to remove a selected file (via some type of delete button). That way if I upload say a previous logfile but say I don't want that anymore, I can delete it without having to refresh the app.
> * I think upload buttons should look a little different. My suggestion would be to have just a rendering of the filename with actual upload/delete buttons as needed. Here's an example of what that might look like (cc'ing @Dtphelan1):
>   image

So this could be my bad for attaching a mock up. I was more so just trying to say this should be included in tandem with something that was already there. Now, we have lost indication of which file box is for the config file, and which one is for the log file. We should add these labels back so the file pickers clearly indicate which kind of file you are uploading for

Also, I realized it's a bit jarring when you upload a file and the path expands and pushes the buttons all the way over to the right. We should figure out a styling where the buttons stay put even when you upload something. One way is to put the buttons over to the left and the file path over to the right. Another is to put the filepath either above or below the buttons so the buttons don't move. Which option do you think is best?

react-app/src/stylesheets/Page.scss
```diff
@@ -75,3 +90,27 @@
...
+
+.file-button-container {
+  display: flex;
+  justify-content: space-betweeen;
+}
+
+.file-picker {
+  font-size: 14px;
+  margin-bottom: 0px;
+  margin-left: 20px;
```
Dtphelan1, line 109: DP3 - This style is used on more than just the file-picker button; it's also used on the delete button. Could we rename the class to be something more appropriate? Also I'll highlight again the utility classes in the Bootstrap cheatsheet that allow for easy inline margin-manipulation (see the `mb-x` classes [here](https://hackerthemes.com/bootstrap-cheatsheet/)). Those may be of use in lieu of a specific class. Up to you.

Dtphelan1:
Last thing: For the file uploaders, I think it could be nicer to have the buttons to the left and the string to the right. Could you update the order of those?

Dtphelan1 (approved):
There are a few small styling things that we could continue playing with, but I think this looks good to me. Great work Seneca!

mgramigna (approved):

PR #8, opened 2021-06-29
Resource counts

SenecaAnderson:
### Summary
When you click on the "Patient x" dropdowns in the sidebar of the output display, the accordion expands and displays a list of resources and the number of each. The total number of resources is at the bottom of the list.

### Code Changes
Several functions were added to the Result component, and the JSX returned by it was modified. A CSS class was added to Pages.scss to make the "Total Resources: x" have more visible styling.

### New Behavior
When you click on the "Patient x" dropdowns, a list of resources and the number of each appears.

### Testing Guidance
Use npm start to start the app. Click on "Extract New". Provide a valid config file (as well as anything else you'd like), then click "Submit". Click on the "Patient x" dropdowns in the sidebar to toggle between patients and view the resource counts.

### DIFF
--- a/react-app/src/components/Result.js
+++ b/react-app/src/components/Result.js
@@ -5,7 +5,7 @@ const _ = require('lodash');
 const fhirpath = require('fhirpath');
 
 function Result(props) {
-  function getResourceList() {
+  function countResources() {
     const resourceList = [];
     // iterate through resources array, checking the resourceType of each entry
     // if (new type),
@@ -13,25 +13,31 @@ function Result(props) {
     const allResourceTypesPath = 'Bundle.descendants().resource.resourceType';
     const uniqueResourceTypes = _.uniq(fhirpath.evaluate(props.bundle, allResourceTypesPath));
 
-    let totalResources = 0;
-    // FIXME - sort uniqueResourceTypes alphabetically
-
     const countThisResource = (resourceType) =>
       `Bundle.descendants().where(resource.resourceType = '${resourceType}').count()`;
     uniqueResourceTypes.forEach((resourceType) => {
       const count = fhirpath.evaluate(props.bundle, countThisResource(resourceType));
       resourceList.push({ resourceType, count });
-      totalResources += parseInt(count, 10);
     });
 
-    let sortedList = [{ resourceType: 'Total Resources', count: totalResources }];
-    sortedList = sortedList.concat(_.sortBy(resourceList, ['resourceType', 'count']));
+    return _.sortBy(resourceList, ['resourceType', 'count']);
+  }
 
-    return sortedList.map((item, i) => (
+  function getResourceList() {
+    const resourceList = countResources();
+    return resourceList.map((item, i) => (
       <p key={i}>
         {item.resourceType}: {item.count}
       </p>
     ));
+  }
+
+  function getNumResources() {
+    let total = 0;
+    countResources().forEach((item) => {
+      total += parseInt(item.count, 10);
+    });
+    return total;
   }
 
   return (
@@ -44,7 +50,10 @@ function Result(props) {
       >
         Patient {props.id + 1}
       </Accordion.Header>
-      <Accordion.Body>{getResourceList()}</Accordion.Body>
+      <Accordion.Body>
+        {getResourceList()}
+        <p className="emphasized-list-text">Total Resources: {getNumResources()}</p>
+      </Accordion.Body>
     </Accordion.Item>
   );
 }
--- a/react-app/src/components/Result.js
+++ b/react-app/src/components/Result.js
@@ -10,10 +10,11 @@ function Result(props) {
     const allResourceTypesPath = 'Bundle.descendants().resource.resourceType';
     const uniqueResourceTypes = _.uniq(fhirpath.evaluate(props.bundle, allResourceTypesPath));
 
-    const countThisResource = (resourceType) =>
-      `Bundle.descendants().where(resource.resourceType = '${resourceType}').count()`;
     uniqueResourceTypes.forEach((resourceType) => {
-      const count = fhirpath.evaluate(props.bundle, countThisResource(resourceType));
+      const count = fhirpath.evaluate(
+        props.bundle,
+        `Bundle.descendants().where(resource.resourceType = '${resourceType}').count()`,
+      );
       resourceList.push({ resourceType, count });
     });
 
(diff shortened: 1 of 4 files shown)


react-app/src/components/Result.js
```diff
@@ -1,7 +1,45 @@
...
 import { Accordion } from 'react-bootstrap';
 
+const _ = require('lodash');
+const fhirpath = require('fhirpath');
+
 function Result(props) {
+  function countResources() {
+    const resourceList = [];
+    // iterate through resources array, checking the resourceType of each entry
+    // if (new type),
```
Dtphelan1, line 11: DP1 - We could probably drop this right?

react-app/src/components/Result.js
```diff
@@ -1,7 +1,45 @@
...
+
+    const allResourceTypesPath = 'Bundle.descendants().resource.resourceType';
+    const uniqueResourceTypes = _.uniq(fhirpath.evaluate(props.bundle, allResourceTypesPath));
+
+    const countThisResource = (resourceType) =>
```
Dtphelan1, line 16: DP2 - Does this need to live above the forEach loop? Could we just create this string within the body of the loop?  It's was a little confusing that we have a function that returns a string instead of just having a string that uses string interpolation in the forEach loop

react-app/src/components/Result.js
```diff
@@ -1,7 +1,45 @@
...
+      resourceList.push({ resourceType, count });
+    });
+
+    return _.sortBy(resourceList, ['resourceType', 'count']);
+  }
+
+  function getResourceList() {
+    const resourceList = countResources();
+    return resourceList.map((item, i) => (
+      <p key={i}>
```
Dtphelan1, line 29: DP3 - It's discouraged to use index-in-an-array as a key for a `.map`. The reasons for it being less-than-ideal won't materialize here, but we should really be using something that is unique to the thing being rendered in the iteration of the loop (for example, the resourceType!). If you want to read more look here: https://robinpokorny.medium.com/index-as-a-key-is-an-anti-pattern-e0349aece318

react-app/src/components/Result.js
```diff
@@ -10,8 +48,12 @@ function Result(props) {
           props.setShowLogs(false);
         }}
       >
-        Patient {props.id}
+        Patient {props.id + 1}
       </Accordion.Header>
+      <Accordion.Body>
+        {getResourceList()}
+        <p className="emphasized-list-text">Total Resources: {getNumResources()}</p>
```
Dtphelan1, line 55: DP4 - I think I'd prefer Total Resources above the count-by-each-resource-type

Dtphelan1 (changes requested):
Let me know when those changes are done and I'll take another look!

react-app/src/components/Result.js
SenecaAnderson: This is function is a modified version of the getResourceCountInBundle function from /mcode-extraction-framework/src/helpers/fhirUtils.js. I deleted it from my code and it works fine. If you want I could make the same change in the original function and submit a pull request.

Dtphelan1: I don't think you need to worry about making the change over there, making the change here is sufficient for this task.

Dtphelan1 (approved):
This looks good to me count-wise. Even after we get a second approval here, we will want to hold-off merge until the other PR is merged in, and then change the base branch here to be `master`.

jafeltra (commented):
I just had a few comments, but this looks good!

react-app/src/components/Result.js
```diff
@@ -10,8 +46,12 @@ function Result(props) {
           props.setShowLogs(false);
         }}
       >
-        Patient {props.id}
+        Patient {props.id + 1}
```
jafeltra, line 49: I have one more question for you and @Dtphelan1. I think previously we have been trying to keep from assuming that the id we display will be an integer, so I'm not sure if we should be incrementing this value here. Was there a reason you decided to change it here? It seems like it might be slightly more complicated to change the id we pass in based on how it gets used in other places, so before we decide to change things, I'm curious if there was a reason for the change and what you and Dylan both think about displaying a `Patient 0` vs `Patient 1` for the first patient.

SenecaAnderson: The id that's displayed is the iterator variable from here, so it's guaranteed to be an int:
`const list = props.extractedData.map((bundle, i) => (
    <Result bundle={bundle} id={i} setPatientID={props.setPatientID} key={i} setShowLogs={props.setShowLogs} />
  ));`

In ResultPage, patientID (which is modified whenever someone clicks on a patient) is used as an index variable to determine which JSON schema gets passed to PatientData:
`<PatientData
              patientJson={props.extractedData[patientID]}
              id={patientID}
              showLogs={showLogs}
              loggedMessages={props.loggedMessages}
            /><PatientData
              patientJson={props.extractedData[patientID]}
              id={patientID}
              showLogs={showLogs}
              loggedMessages={props.loggedMessages}
            />`

Since it's generated by calling extractedData.map() used to access bundles in the extractedData array, the count starts at 0. However, I thought starting with 0 in the output might seem strange/confusing to people who don't have a computer science background.

SenecaAnderson: The 'Patient X' thing was also originally intended to be a placeholder for some more useful title. If you want, I could use fhirpath to retrieve the patient's name or mrn id, and display that instead.

jafeltra: I definitely agree with you on both points - right now the id is guaranteed to be an integer and it would probably be nice to not index a list by zero. However, I think [these comments on previous PRs](https://github.com/mcode/mcode-extraction-ui/pull/3#discussion_r653012115) also discuss how even though we are only using an integer for right now, we want to make sure we aren't locked into using an integer id. This would be another example of when we want to be a little more flexible around how we develop this component to help us ensure we can support different types of data in the future.

I do really like the idea of using `fhirpath` to retrieve the name or another piece of information here. We can discuss at stand up (which is right now haha) about if we want to do that now or if we want to revert this line and do the fhirpath piece separately.

jafeltra: Following up on the conversation we had at stand up, we'll plan to change back the display here to just be `props.id` so we avoid requiring an integer, but I'll make a task to explore what might the best way to label each patient in the side bar.

jafeltra (approved):
This looks good to me! We'll just have to rebase this once the date picker branch is merged.

jafeltra:
I know I said we will need to rebase, but after talking to @Dtphelan1 I think the rebase might not work well since we have so many branches off of each other. We probably should be using a merge instead of a rebase, but we can discuss as a group when it's time to update.

PR #9, opened 2021-06-30
Logger statistics

SenecaAnderson:
Summary
Logger statistics display in the "Log File" dropdown in the sidebar of the result page.

Code Changes
- Changed return value of InMemoryTransport
- Changed LogList to account for new return value
- Modified sidebar to calculate number of errors, warnings, debug messages, and total messages

New Behavior
When you click on the Log File dropdown in the sidebar of the result page, the number of each type of message (excluding info-level messages) will appear in a list.

Testing Guidance
Run the app with npm start. Run extraction with any input that doesn't cause fatal errors. Click on "Log File" in the sidebar. The message counts will appear underneath.

### DIFF
--- a/electron/InMemoryTransport.js
+++ b/electron/InMemoryTransport.js
@@ -11,7 +11,9 @@ module.exports = class InMemoryTransport extends Transport {
     const MESSAGE_SYMBOL = Symbol.for('message');
     const message = info[MESSAGE_SYMBOL];
     const strippedMessage = stripAnsi(message);
-    this.loggedMessages.push(strippedMessage);
+    console.log('info: ', info);
+    console.log('level: ', stripAnsi(info.level));
+    this.loggedMessages.push({ message: strippedMessage, level: stripAnsi(info.level) });
     // Perform the writing to the remote service
     callback();
   }
--- a/electron/InMemoryTransport.js
+++ b/electron/InMemoryTransport.js
@@ -11,8 +11,6 @@ module.exports = class InMemoryTransport extends Transport {
     const MESSAGE_SYMBOL = Symbol.for('message');
     const message = info[MESSAGE_SYMBOL];
     const strippedMessage = stripAnsi(message);
-    console.log('info: ', info);
-    console.log('level: ', stripAnsi(info.level));
     this.loggedMessages.push({ message: strippedMessage, level: stripAnsi(info.level) });
     // Perform the writing to the remote service
     callback();
--- a/react-app/src/components/ResultSidebar.js
+++ b/react-app/src/components/ResultSidebar.js
@@ -15,6 +15,37 @@ function ResultSidebar(props) {
     // Save the results permanently somehow
   }
 
+  function getLoggerStats() {
+    let errors = 0;
+    let warnings = 0;
+    let debugs = 0;
+    let total = 0;
+
+    props.loggedMessages.forEach((log) => {
+      if (log.level === 'error') {
+        errors += 1;
+      }
+      if (log.level === 'warn') {
+        warnings += 1;
+      }
+      if (log.level === 'debug') {
+        debugs += 1;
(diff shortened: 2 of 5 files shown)


electron/InMemoryTransport.js
```diff
@@ -11,7 +11,9 @@ module.exports = class InMemoryTransport extends Transport {
     const MESSAGE_SYMBOL = Symbol.for('message');
     const message = info[MESSAGE_SYMBOL];
     const strippedMessage = stripAnsi(message);
-    this.loggedMessages.push(strippedMessage);
+    console.log('info: ', info);
+    console.log('level: ', stripAnsi(info.level));
```
jafeltra, line 15: I'm guessing you were using these `console.log`s to develop/debug, but we should remove them now.

react-app/src/components/ResultSidebar.js
```diff
@@ -15,6 +15,37 @@ function ResultSidebar(props) {
...
+        debugs += 1;
+      }
+      total += 1;
+    });
+
+    return (
+      <Accordion.Body>
+        <p key={0}>Errors: {errors}</p>
+        <p key={1}>Warnings: {warnings}</p>
+        <p key={2}>Debug Messages: {debugs}</p>
```
jafeltra, line 41: I don't think we need to report the number of debug messages. The debug messages just provide additional detail, but the number of them isn't really indicative of the number of problems that will require attention, like the error and warn messages.

react-app/src/components/ResultSidebar.js
```diff
@@ -15,6 +15,37 @@ function ResultSidebar(props) {
...
+      if (log.level === 'debug') {
+        debugs += 1;
+      }
+      total += 1;
+    });
+
+    return (
+      <Accordion.Body>
+        <p key={0}>Errors: {errors}</p>
+        <p key={1}>Warnings: {warnings}</p>
```
jafeltra, line 40: Since these `<p>` elements (and the ones below) are not in a loop, I don't think we need to have the `key` values on them. There should be no warnings in the console when you remove them.

react-app/src/components/ResultSidebar.js
```diff
@@ -15,6 +15,37 @@ function ResultSidebar(props) {
...
+        <p key={2}>Debug Messages: {debugs}</p>
+        <p className="emphasized-list-text" key={3}>
+          Total Messages: {total}
```
jafeltra, line 43: Since the total count of resources was moved to the top, we should move the total messages count to match.

jafeltra (changes requested):
This is looking good. I just had a few comments inline. Let me know if you have any questions.

react-app/src/components/ResultSidebar.js
SenecaAnderson: I've moved the total message count. It uses the same CSS class as the total resource count text, so the color will update during rebasing.

react-app/src/components/ResultSidebar.js
```diff
@@ -28,20 +27,14 @@ function ResultSidebar(props) {
...
 
     return (
       <Accordion.Body>
-        <p key={0}>Errors: {errors}</p>
-        <p key={1}>Warnings: {warnings}</p>
-        <p key={2}>Debug Messages: {debugs}</p>
-        <p className="emphasized-list-text" key={3}>
-          Total Messages: {total}
-        </p>
+        <p className="emphasized-list-text">Total Messages: {total}</p>
```
jafeltra, line 35: Just writing down what we talked about in the working session, but I think we can remove the total count and just keep the total error and warning messages.

jafeltra (approved):
This looks good to me!

Dtphelan1 (approved):
This looks great. Still holding off on this until the other PR's are merged, but well done

PR #10, opened 2021-07-01
Save output

SenecaAnderson:
### Summary
Users can save the patient JSON schemas to a folder on their computer.

### New Behavior
When the user clicks the "Save" button on the output page, a file explorer opens up, allowing them to select a folder to download the patient JSON schemas to. The page will display a success alert if saving succeeded, and an error alert if fs.writeFileSync() throws an error.

### Code Changes
- Functions added to preload.js and electron.js to handle the backend of opening the file explorer and writing files
- onSave() function now has code in it
- Alerts that display when after the user saves (or fails to save) info

### Testing Guidance
Start the app with npm start. Click "Extract New". Run extraction with any set of data that will not lead to fatal errors. Click "Save" and choose a folder. 
- If there is patient data to save, it will save, and an alert will appear informing you that it has saved. Click the "x" to dismiss the alert.
- If saving fails, an error alert will appear. Click the "x" to dismiss it.
- If you cancel without choosing a folder, there will be no alert.
- If you try to save when there's no patient data (reload the window to remove patient data), a warning will appear informing you that there's nothing to save. Click the "x" to dismiss it.

### DIFF
--- a/electron/electron.js
+++ b/electron/electron.js
@@ -77,6 +77,7 @@ ipcMain.handle('get-file', async () =>
 
 ipcMain.handle('get-output-path', async () => {
   const options = {
+    buttonLabel: 'Save',
     title: 'Select Output Folder',
     defaultPath: app.getPath('downloads'),
     properties: ['openDirectory', 'createDirectory'],
--- a/electron/electron.js
+++ b/electron/electron.js
@@ -85,8 +85,14 @@ ipcMain.handle('get-output-path', async () => {
 });
 
 ipcMain.handle('save-output', async (event, savePath, extractedData) => {
-  extractedData.forEach((bundle, i) => {
-    const outputFile = path.join(savePath, `mcode-extraction-patient-${i + 1}.json`);
-    fs.writeFileSync(outputFile, JSON.stringify(bundle), 'utf8');
-  });
+  try {
+    extractedData.forEach((bundle, i) => {
+      const outputFile = path.join(savePath, `mcode-extraction-patient-${i + 1}.json`);
+      fs.writeFileSync(outputFile, JSON.stringify(bundle), 'utf8');
+      // 'false' as in 'no error'
+    });
+    return true;
+  } catch (error) {
+    return false;
+  }
 });
--- a/electron/electron.js
+++ b/electron/electron.js
@@ -90,8 +90,10 @@ ipcMain.handle('save-output', async (event, savePath, extractedData) => {
       const outputFile = path.join(savePath, `mcode-extraction-patient-${i + 1}.json`);
       fs.writeFileSync(outputFile, JSON.stringify(bundle), 'utf8');
     });
+    // retuerning true indicates that the save process succeeded
     return true;
   } catch (error) {
+    // returning false indicates that something went wrong during the save process
     return false;
   }
 });
--- a/electron/electron.js
+++ b/electron/electron.js
@@ -89,7 +89,6 @@ ipcMain.handle('save-output', async (event, savePath, extractedData) => {
     extractedData.forEach((bundle, i) => {
       const outputFile = path.join(savePath, `mcode-extraction-patient-${i + 1}.json`);
       fs.writeFileSync(outputFile, JSON.stringify(bundle), 'utf8');
-      // 'false' as in 'no error'
     });
     return true;
   } catch (error) {
--- a/electron/electron.js
+++ b/electron/electron.js
@@ -93,7 +93,7 @@ ipcMain.handle('save-output', async (event, savePath, extractedData) => {
     // retuerning true indicates that the save process succeeded
     return true;
   } catch (error) {
-    // returning false indicates that something went wrong during the save process
-    return false;
+    // return the error message
+    return error.message;
   }
 });
--- a/electron/electron.js
+++ b/electron/electron.js
@@ -1,6 +1,6 @@
-const { app, BrowserWindow, ipcMain } = require('electron');
-const { dialog } = require('electron');
+const { app, BrowserWindow, dialog, ipcMain } = require('electron');
 const squirrel = require('electron-squirrel-startup');
+const fs = require('fs');
 const path = require('path');
 const { logger } = require('mcode-extraction-framework');
 const runExtraction = require('./extraction');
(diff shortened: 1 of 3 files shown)


electron/electron.js
```diff
@@ -74,3 +74,25 @@ ipcMain.handle('get-file', async () =>
...
   }),
 );
+
+ipcMain.handle('get-output-path', async () => {
+  const options = {
+    title: 'Select Output Folder',
+    defaultPath: app.getPath('downloads'),
+    properties: ['openDirectory', 'createDirectory'],
+  };
+  return dialog.showOpenDialog(null, options, (savePath) => savePath);
```
Dtphelan1, line 84: DP1 - Take a look at the dialog documentation to see if there is a better alternative to this function here. https://www.electronjs.org/docs/api/dialog

electron/electron.js
```diff
@@ -74,3 +74,25 @@ ipcMain.handle('get-file', async () =>
...
+});
+
+ipcMain.handle('save-output', async (event, savePath, extractedData) => {
+  try {
+    extractedData.forEach((bundle, i) => {
+      const outputFile = path.join(savePath, `mcode-extraction-patient-${i + 1}.json`);
+      fs.writeFileSync(outputFile, JSON.stringify(bundle), 'utf8');
+      // 'false' as in 'no error'
```
Dtphelan1, line 92: DP2 - This placement and text feels outdated. Can we update this comment to be closer to what it's referring to and to be more accurate?

electron/electron.js
SenecaAnderson: What different features should I look for in an alternative?

react-app/src/components/ResultSidebar.js
```diff
@@ -1,18 +1,35 @@
...
-    // Save the results permanently somehow
+    if (props.extractedData.length > 0) {
+      window.api.getOutputPath().then((savePath) => {
+        if (!savePath.canceled) {
+          const result = window.api.saveOutput(savePath.filePaths[0], props.extractedData);
+          if (result) {
+            setShowSavedAlert(true);
+          } else {
+            setShowErrorAlert(true);
+          }
```
mgramigna, line 27: I don't think this block is doing what it appears. The `saveOutput` function is an `async` function that eventually resolves to a boolean value from the `save-output` IPC message. You either need to use `await` or chain another `.then` (do not nest `.then`s inside each other, this is an anti pattern. Read up on promise chaining here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises#chaining)

Right now, `result` is actually an unresolved promise that _might_ be `true` or `false`, but it will always pass the first `if` check because the Promise is never awaited or `.then`ed

mgramigna (changes requested):

electron/electron.js
```diff
@@ -74,3 +74,24 @@ ipcMain.handle('get-file', async () =>
...
+  try {
+    extractedData.forEach((bundle, i) => {
+      const outputFile = path.join(savePath, `mcode-extraction-patient-${i + 1}.json`);
+      fs.writeFileSync(outputFile, JSON.stringify(bundle), 'utf8');
+    });
+    return true;
+  } catch (error) {
+    return false;
+  }
+});
```
mgramigna, line 97: I think the flow can probably be improved here such that in the error cases, we send some time of error message back to the client so the user is aware that an issue happened. 

Perhaps this return value can be changed such that `error.message` makes its way back to the react app? And then instead of just rendering "output saved" or not, we could render the error message in the error case

electron/electron.js
Dtphelan1: Well this is a dialog for opening a file. Explore the documentation to see if we can find one that's more applicable to what we're actually doing with that dialog, instead of opening a file.

electron/electron.js
Dtphelan1: @SenecaAnderson to be clear, I didn't ask you to remove the comment. Could we add an updated comment that sensibly communicates what false and true mean in this context?

electron/electron.js
SenecaAnderson: The other option is dialog.showSaveDialog, but unfortunately it turns out that showSaveDialog doesn't have the option of just selecting a folder--the dialog requires the user to input a file name to save as. Since the save process is supposed to save multiple files with predetermined names to a folder, the showSaveDialog function doesn't match with the purpose of the save process, and may confuse users. The dialog.showOpenDialog function has an option ('openDirectory') that lets you limit it so that the user can only select folders. 
Luckily it turns out there's an option to change the confirmation button text to something custom. I changed it from "Open" to "Save", so now showOpenDialog looks the same as showSaveDialog, only without the section to input a file name to save as. Is this an okay solution, or should I keep looking for another way?

react-app/src/components/ResultSidebar.js
SenecaAnderson: Thanks for the resource on promise chaining! That clarified a lot of things

mgramigna:
It's possible that something came up on a previous PR for this, but when trying to test the save button, I noticed that expanding the panel for resource counts can push down the buttons on small screens so they are out of view:

image

I think this should be added to the backlog as a relatively low priority task because it's niche behavior and only happens on certain window sizes. Just wanted to note that here

mgramigna:
Waiting for resolution on the save/open dialog discussion, but rest of this looks good to me

Dtphelan1:
Did we make a task for MG's comment above?

mgramigna:
> Did we make a task for MG's comment above?

on it

electron/electron.js
Dtphelan1: I don't think you've pushed that change yet, but if that looks okay to you I'd be fine with that for now – primarily because we're looking to implement a more sophisticated UI for saving all extracted data. Also let me know if you're still looking into a custom UI for this. If you've put that on pause, just let me know.

Dtphelan1:
Want to know the status of your changes on the custom dialog, but after that I'll reevaluate approving.

SenecaAnderson:
I think PR #12 (Scrolling Sidebar) might already address MG's comment, so we might not need a new task

electron/electron.js
SenecaAnderson: Okay. I've pushed the button label change. I'm still looking into using a more sophisticated UI for this, or potentially a form inside of a showMessageBox.

mgramigna (approved):

SenecaAnderson:
I have addressed all the current feedback on this branch.

Dtphelan1 (approved):
looks good!

PR #13, opened 2021-07-07
Patient result label

SenecaAnderson:
### Summary
The patient label in the sidebar of the result page now changes based on available information in the JSON schema.

### New Behavior
Previously, the patients were all labelled "Patient X", where X was an iterator variable used to access the patient's bundle in extractedData. Now, the label changes based on available information:
1. If both an MRN and a name are available, they are displayed in this format: "MRN: First Last"
2. If either an MRN or a name is available, but not both, than the available data displays by itself
3. If neither an MRN nor a name is available, the patient resource ID displays
4. If a resource ID is unavailable, the old "Patient X" label is used.

### Code Changes
- Added function to Result component and modified JSX in return statement
- Changes to CSS classes so that the resource IDs (which are very long strings) don't overflow out of the sidebar. In order to get this working properly, I also made the sidebar a set width--previously, it only had a minimum width, and would expand at large window sizes, which looked odd.  

### Testing Guidance
1. Start the app with npm start. Click on "Extract New".
2. In patient-information.csv, delete the first and last name of one of the patients. Make sure that there is at least one other patient with their name available.
3. Upload this config file and run extraction. Look at the sidebar of the result page. The patients with names available should be labelled "MRN: First Last". The patient with no name should be labelled "MRN". If there is a patient with no data, they will be labelled "Patient X".
4. Add lines to the patient extractor in the configuration file in order to mask the MRNs:
`{
      "label": "patient",
      "type": "CSVPatientExtractor",
      "constructorArgs": {
        "filePath": "./sample-extraction-data/sample-client-data/patient-information.csv",
        "mask": [
          "mrn"
        ]
      }
    },
5. Click "Exit" in the result page. Upload the config file again and click "Submit". Look at the sidebar of the result page. The patients with names available will be labelled "First Last". The patient with no name will be labelled with the patient resource ID. If there's an empty patient, they will be labelled "Patient X".

### DIFF
--- a/react-app/src/components/Result.js
+++ b/react-app/src/components/Result.js
@@ -43,36 +43,27 @@ function Result(props) {
       const mrn = patient.id;
       const name = patient.name[0].text;
 
-      let isMasked = fhirpath.evaluate(patient, 'Patient.identifier.extension.valueCode')[0];
-
-      if (isMasked === 'masked') {
-        isMasked = true;
-      } else {
-        isMasked = false;
-      }
+      const isMasked = fhirpath.evaluate(patient, 'Patient.identifier.extension.valueCode')[0] === 'masked';
 
       // if both MRN and name -- return string w / both
       if (typeof mrn === 'string' && mrn.length > 0 && typeof name === 'string' && name.length > 0 && !isMasked) {
         const label = mrn.concat(': ').concat(name);
         return <p className="accordion-result-label">{label}</p>;
       }
+      // if either MRN or name -- return string / the available one
       if (typeof name === 'string' && name.length > 0) {
         return <p className="accordion-result-label">{name}</p>;
       }
+      // if neither MRN nor name -- "Patient " + patient_resource_id
       if (typeof mrn === 'string' && mrn.length > 0) {
         return <p className="accordion-result-label">{mrn}</p>;
       }
     }
 
+    // if no patient resource ID -- return "Patient " + props.id
     let label = 'Patient';
     label = label.concat(' ').concat(props.id.toString());
     return <p className="accordion-result-label">{label}</p>;
-
-    // if either MRN or name -- return string / the available one
-
-    // if neither MRN nor name -- "Patient " + patient_resource_id
-
-    // if no patient resource ID -- return "Patient " + props.id
   }
 
   const resourceList = countResources();
--- a/react-app/src/components/Result.js
+++ b/react-app/src/components/Result.js
@@ -36,6 +36,20 @@ function Result(props) {
     return total;
   }
 
+  function getLabel() {
+    // attempt to get MRN and name with fhirpath
+    const mrnPath = 'Bundle.descendants().resource.resourceType';
+    const namePath = 'Bundle.descendants().resource.resourceType';
+
+    // if both MRN and name -- return string w / both
+
+    // if either MRN or name -- return string / the available one
+
+    // if neither MRN nor name -- "Patient " + patient_resource_id
+
+    // if no patient resource ID -- return "Patient " + props.id
+  }
+
   const resourceList = countResources();
 
   return (
--- a/react-app/src/components/Result.js
+++ b/react-app/src/components/Result.js
@@ -51,20 +51,20 @@ function Result(props) {
     // if both MRN and name -- return string w / both
     if (typeof mrn === 'string' && mrn.length > 0 && typeof name === 'string' && name.length > 0) {
       const label = mrn.concat(': ').concat(name);
-      return label;
+      return <p className="accordion-result-label">{label}</p>;
     }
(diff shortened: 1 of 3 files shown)


react-app/src/components/ResultPage.js
```diff
@@ -14,7 +14,7 @@ function ResultPage(props) {
     <div>
       <Container fluid>
         <Row>
-          <Col xxl={3} xl={3} lg={3} md={3} sm={3} xs={3} className="sidebar-col">
+          <Col xxl={1} xl={1} lg={1} md={1} sm={1} xs={3} className="sidebar-col">
```
Dtphelan1, line 17: DP - Could you look at the grid system documentation closer and see if there is a way to reduce the number of breakpoints we use here?

react-app/src/components/Result.js
```diff
@@ -36,6 +36,48 @@ function Result(props) {
...
+      props.bundle,
+      "Bundle.descendants().resource.where(resourceType='Patient').name.text",
+    )[0];
+
+    let isMasked = fhirpath.evaluate(
+      props.bundle,
+      "Bundle.descendants().resource.where(resourceType='Patient').identifier.extension.valueCode",
+    )[0];
+
+    if (typeof isMasked === 'string' && isMasked === 'masked') {
```
Dtphelan1, line 52: DP - Do we need this typeof check since we're using triple-equals?

react-app/src/components/ResultPage.js
SenecaAnderson: It looks like I can delete the xxl, xl, lg, and md breakpoints and have the same effect

react-app/src/components/Result.js
```diff
@@ -36,6 +36,48 @@ function Result(props) {
...
+    const mrn = fhirpath.evaluate(props.bundle, "Bundle.descendants().resource.where(resourceType='Patient').id")[0];
+    const name = fhirpath.evaluate(
+      props.bundle,
+      "Bundle.descendants().resource.where(resourceType='Patient').name.text",
+    )[0];
+
+    let isMasked = fhirpath.evaluate(
+      props.bundle,
+      "Bundle.descendants().resource.where(resourceType='Patient').identifier.extension.valueCode",
+    )[0];
```
mgramigna, line 50: I think this could use a slight refactor to avoid 3 very similar fhirpath evaluates on the entire bundle. Notice the common chunk of the fhirpath expression that you're using to get the Patient resource. I'd suggest an approach where you use one evaluate to grab that patient resource. Then some of the other things should be trivial:

* Imagine you already have a patient resource when trying to access `.name.text`: do you even need fhirpath?
* The first argument to `fhirpath.evaluate` can be any json object, can you think of a way to do the `isMasked` logic with `fhirpath` but in a way that avoids repeating the chunk to find the patient? (assuming you already have a reference to the patient resource in the bundle at this point)

react-app/src/components/Result.js
```diff
@@ -36,6 +36,45 @@ function Result(props) {
...
+
+    let label = 'Patient';
+    label = label.concat(' ').concat(props.id.toString());
+    return <p className="accordion-result-label">{label}</p>;
+
+    // if either MRN or name -- return string / the available one
+
+    // if neither MRN nor name -- "Patient " + patient_resource_id
+
+    // if no patient resource ID -- return "Patient " + props.id
```
Dtphelan1, line 75: Could we reorganize these comments to wherever they should be?

react-app/src/components/Result.js
```diff
@@ -36,6 +36,45 @@ function Result(props) {
...
+      const mrn = patient.id;
+      const name = patient.name[0].text;
+
+      let isMasked = fhirpath.evaluate(patient, 'Patient.identifier.extension.valueCode')[0];
+
+      if (isMasked === 'masked') {
+        isMasked = true;
+      } else {
+        isMasked = false;
+      }
```
Dtphelan1, line 52: Surely we could define this with a `const` without the conditional here, right?

react-app/src/components/Result.js
SenecaAnderson: I have refactored this code.

react-app/src/components/Result.js
SenecaAnderson: I have moved these comments.

react-app/src/components/Result.js
SenecaAnderson: I have modified this so that it's a const and only one line.

SenecaAnderson:
I have addressed all current feedback on this branch.

Dtphelan1 (approved):
This looks good to me!

Dtphelan1:
I think this needs one more glance by @mgramigna but other than that it should be good 😄

mgramigna (approved):

PR #14, opened 2021-07-07
Color coded log messages based on levels

SenecaAnderson:
### Summary
The log messages are now color coded. The level is inside a badge, and the color of the badge changes based on the level.

### New Behavior
The log messages are color-coded.

### Code Changes
- Split up the message, level, and timestamp in the object returned by InMemoryTransport
- Modified LogList JSX to display the Badge with the level inside, the timestamp, and then the message
- Added function to LogList to determine which variant to use in Badge
- Added fixed-with-40px CSS class to LogList.scss so that the badges are all the same width

### Testing guidance
Start the app with npm start. Click on "Extract New". Run extraction with any data (both the error page and result page display the LogList component). If you use valid data, click on "Log File" in the sidebar of the result page. The LogList component will display.

### DIFF
--- a/react-app/src/components/LogList.js
+++ b/react-app/src/components/LogList.js
@@ -1,12 +1,31 @@
 import React from 'react';
-import { ListGroup } from 'react-bootstrap';
+import { Badge, ListGroup } from 'react-bootstrap';
 
 import '../stylesheets/LogList.scss';
 
 function LogList(props) {
+  function getVariant(level) {
+    if (level === 'info') {
+      return 'success';
+    }
+    if (level === 'warn') {
+      return 'warning';
+    }
+    if (level === 'error') {
+      return 'danger';
+    }
+    if (level === 'debug') {
+      return 'primary';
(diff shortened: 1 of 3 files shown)


dmendelowitz (approved):
This looks great! Side note: I think I'm the first Windows user to test the UI, so I can confirm that everything you've done up to this point works on Windows as well

react-app/src/components/LogList.js
```diff
@@ -1,12 +1,31 @@
...
+      return 'warning';
+    }
+    if (level === 'error') {
+      return 'danger';
+    }
+    if (level === 'debug') {
+      return 'primary';
+    }
+    return 'primary';
+  }
```
Dtphelan1, line 21: DP1 - Could we use a switch here?

SenecaAnderson:
I have addressed all of the current feedback on the branch.

Dtphelan1 (approved):
This looks good to me

PR #16, opened 2021-07-16
Add Save As functionality and back buttons to config editor

SenecaAnderson:
### Summary
The config editor now has functional "Back" and "Save As" buttons. Additionally, I found a fix for the bug where Buttons in LinkContainers always have active styling.

### New Behavior
The config menu has a back button that redirects to the home page. The config editor has a back button that toggles back to the config menu. The config editor has a "Save As" button which allows the user to save the config schema. If the save process succeeds, a green success alert appears. If the save process fails, a red error alert appears and displays the error message. If the user cancels the save process, the save dialog disappears and no alert displays.

### Code Changes
- Added "Back" button to config menu
- Added "Back" and "Save As" buttons to ConfigForm, along with associated functions
- Added save-config-as functions to backend (preload.js & electron.js)
- Added button-text CSS style in order to get around the LinkContainer active styling bug
- Added schema variable to ConfigEditor. The variable currently contains a placeholder JSON schema for testing purposes.

### Testing Guidance
Start the app with npm start. Click "Configuration File Editor". Click "Back". You will return to the home page. Click on "Configuration File Editor" again. Click on "Create New". The ConfigForm component will appear. Click "Back" to go back to the config menu. Click on "Save As" to open the save dialog and save the schema. If it succeeds, a green success alert will appear. If it fails, a red error message alert will appear. If you click "Cancel", the save dialog will disappear and nothing will happen in the app.

### DIFF
--- a/electron/electron.js
+++ b/electron/electron.js
@@ -91,10 +91,39 @@ ipcMain.handle('save-output', async (event, savePath, extractedData) => {
       const outputFile = path.join(savePath, `mcode-extraction-patient-${i + 1}.json`);
       fs.writeFileSync(outputFile, JSON.stringify(bundle), 'utf8');
     });
-    // retuerning true indicates that the save process succeeded
+    // returning true indicates that the save process succeeded
     return true;
   } catch (error) {
     // return the error message
+    return error.message;
+  }
+});
+
+ipcMain.handle('save-config-as', async (event, schema) => {
+  try {
+    const options = {
+      defaultPath: app.getPath('downloads'),
+      properties: ['createDirectory'],
+    };
+    return dialog
+      .showSaveDialog(null, options)
+      .then((savePath) => {
+        if (!savePath.canceled) {
+          return savePath.filePath;
+        }
+        return null;
+      })
+      .then((savePath) => {
+        if (savePath !== null) {
+          fs.writeFileSync(savePath, JSON.stringify(schema), 'utf8');
+          // returning true indicates that the save process succeeded
+          return true;
+        }
+        // returning null indicates that the save process was cacelled
+        return savePath;
+      });
+  } catch (error) {
+    // returning a string indicates that an error occurred
     return error.message;
   }
 });
--- a/electron/electron.js
+++ b/electron/electron.js
@@ -100,30 +100,24 @@ ipcMain.handle('save-output', async (event, savePath, extractedData) => {
 });
 
 ipcMain.handle('save-config-as', async (event, schema) => {
-  try {
-    const options = {
-      defaultPath: app.getPath('downloads'),
-      properties: ['createDirectory'],
-    };
-    return dialog
-      .showSaveDialog(null, options)
-      .then((savePath) => {
-        if (!savePath.canceled) {
-          return savePath.filePath;
-        }
-        return null;
-      })
-      .then((savePath) => {
-        if (savePath !== null) {
-          fs.writeFileSync(savePath, JSON.stringify(schema), 'utf8');
-          // returning true indicates that the save process succeeded
-          return true;
-        }
-        // returning null indicates that the save process was cacelled
-        return savePath;
-      });
-  } catch (error) {
-    // returning a string indicates that an error occurred
-    return error.message;
-  }
+  const options = {
+    defaultPath: app.getPath('downloads'),
+    properties: ['createDirectory'],
+  };
+  return dialog
+    .showSaveDialog(null, options)
+    .then((savePath) => {
+      if (!savePath.canceled) {
+        return savePath.filePath;
+      }
+      return null;
+    })
+    .then((savePath) => {
+      if (savePath !== null) {
+        fs.writeFileSync(savePath, JSON.stringify(schema), 'utf8');
+      }
+      // returning a path indicates that the save process succeeded
+      // returning null indicates that the save process was cancelled
+      return savePath;
+    });
 });
--- a/electron/electron.js
+++ b/electron/electron.js
@@ -99,9 +99,10 @@ ipcMain.handle('save-output', async (event, savePath, extractedData) => {
   }
 });
 
-ipcMain.handle('save-config-as', async (event, schema) => {
+ipcMain.handle('save-config-as', async (event, configJSON) => {
   const options = {
     defaultPath: app.getPath('downloads'),
+    filters: [{ name: 'JSON', extensions: ['json'] }],
     properties: ['createDirectory'],
   };
   return dialog
@@ -114,7 +115,7 @@ ipcMain.handle('save-config-as', async (event, schema) => {
     })
     .then((savePath) => {
       if (savePath !== null) {
-        fs.writeFileSync(savePath, JSON.stringify(schema), 'utf8');
+        fs.writeFileSync(savePath, JSON.stringify(configJSON), 'utf8');
       }
       // returning a path indicates that the save process succeeded
       // returning null indicates that the save process was cancelled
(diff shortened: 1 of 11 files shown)


Dtphelan1:
There are a few bugs that I've encountered, and they all seem to boil down to the fact that there's a better way of combining react-router and react-bootstrap than using the custom dependency LinkContainer component – we should be using the `useHistory`[ hook ](https://reactrouter.com/web/api/Hooks/usehistory). If you take a look at that link, you can see how the documentation recommends we combine an HTML button and the custom hook to achieve the behavior we're seeking. 

The tl;dr of my ask is: 
- Remove the react-router-bootstrap dependency
- Create a custom LinkButton component that consumes a link and some text and creates a button that we can use in a few places.
- Replace all previous uses of the LinkContainer/BootstrapButton combo with this component. 

As for the longer version, it would be better explained over a teams call. I'd be happy to hop in the working se4ssion if that would be helpful! 😄

react-app/src/components/ConfigEditor.js
```diff
@@ -1,25 +1,44 @@
...
-    <div>
+    <div className="page-container">
       <h1 className="page-title">Configuration Editor</h1>
       {!showForm && (
         <div className="button-container">
           <Button className="vertical-menu-button" variant="outline-secondary" onClick={toggleForm}>
             Create New
           </Button>
+
+          <Button className="vertical-menu-button" variant="outline-secondary" onClick={toggleForm}>
```
Dtphelan1, line 34: Do we want to toggle the form on click here?

react-app/src/components/ConfigForm.js
```diff
@@ -1,7 +1,59 @@
...
+      .saveConfigAs(props.schema)
+      .then((result) => {
+        if (result === true) {
+          // if saveOutput() returns true, then the save process succeeded
+          setShowSavedAlert(true);
+        } else if (typeof result === 'string') {
+          // if the result is a string, that means the save process returned an error message
+          setErrorMessage(result);
+          setShowErrorAlert(true);
+        }
```
Dtphelan1, line 19: Is there a reason the saveConfigAs function doesn't just always throw an error when one occurs? do we have control over that api? Could we just be throwing an error if an error occurs and catching it in the catch block below?

react-app/src/components/ConfigEditor.js
SenecaAnderson: This was left over from copy/pasting. I've deleted it.

react-app/src/components/ConfigForm.js
SenecaAnderson: I've fixed this.

react-app/src/stylesheets/Page.scss
```diff
@@ -126,6 +126,10 @@
   margin-top: 0px;
 }
 
+.button-text {
+  margin-bottom: 0px;
+}
+
```
Dtphelan1, line 132: I think we can remove this now since we've gone with the LinkButton component right?

electron/preload.js
```diff
@@ -26,4 +26,8 @@ contextBridge.exposeInMainWorld('api', {
     const result = await ipcRenderer.invoke('save-output', savePath, extractedData);
     return result;
   },
+  saveConfigAs: async (schema) => {
+    const result = await ipcRenderer.invoke('save-config-as', schema);
+    return result;
```
Dtphelan1, line 31: We could probably just return the result of ipcRendered.invoke directly, instead of awaiting this fn call.

react-app/src/components/ConfigForm.js
```diff
@@ -1,7 +1,58 @@
...
+        setShowErrorAlert(true);
+        setErrorMessage(error.message);
+        setShowSavedAlert(false);
+      });
+  }
+  function onBack() {
+    props.setShowForm(false);
+  }
+  return (
+    <div className="page-container">
```
Dtphelan1, line 30: I don't know if we want to use the page-container here. The CSS says this container will always be 100vh; since the config form renders below the title of the configEditor page, the result is that the buttons fall off screen. I've included a snapshop that attempts to show this. 
image

react-app/src/stylesheets/Page.scss
SenecaAnderson: I've removed this.

react-app/src/components/ConfigForm.js
SenecaAnderson: I wound up having to use a state variable to control which class was used for the div surrounding the ConfigEditor component. When the form is displayed, the page as a whole needs to have a height 100vh for the spacing to work properly; when there's no form, the 100vh messes up the button placement.

electron/preload.js
SenecaAnderson: I modified this return statement, as well as the return statements in the other window.api functions.

Dtphelan1 (approved):

Dtphelan1:
Great work Seneca 😄

electron/electron.js
```diff
@@ -91,10 +91,33 @@ ipcMain.handle('save-output', async (event, savePath, extractedData) => {
...
-    // retuerning true indicates that the save process succeeded
+    // returning true indicates that the save process succeeded
     return true;
   } catch (error) {
     // return the error message
     return error.message;
   }
 });
+
+ipcMain.handle('save-config-as', async (event, schema) => {
```
jafeltra, line 102: This is pretty small, but I think it will help us understand the code in the future. What will be saved here isn't a JSON schema, but the JSON that someone built in the editor. I think rather than calling this variable `schema`, we should rename it to reflect that it is just JSON data, so maybe something like `config` or `configJSON`.

electron/electron.js
```diff
@@ -91,10 +91,33 @@ ipcMain.handle('save-output', async (event, savePath, extractedData) => {
...
+  const options = {
+    defaultPath: app.getPath('downloads'),
+    properties: ['createDirectory'],
+  };
+  return dialog
+    .showSaveDialog(null, options)
```
jafeltra, line 108: I'm not sure if this is possible since I haven't looked at the show save dialog documentation, but do you know if it is possible to automatically append the `.json` file extension? Every file we download through the config editor will be a JSON file, and I think it could be nice if we default to that option, similar to how Word defaults to a `.docx` or Powerpoint defaults to `.pptx`.

This is what the dialog looks like for Word: 
image

react-app/src/components/ConfigEditor.js
```diff
@@ -2,24 +2,38 @@ import React, { useState } from 'react';
...
       {!showForm && (
         <div className="button-container">
           <Button className="vertical-menu-button" variant="outline-secondary" onClick={toggleForm}>
             Create New
           </Button>
+          <LinkButton className="vertical-menu-button" variant="outline-secondary" text="Back" path="/" />
         </div>
       )}
-      {showForm && <ConfigForm />}
+      {showForm && <ConfigForm schema={schema} setShowForm={setShowForm} />}
```
jafeltra, line 36: Similar to above, I think the prop `schema` and the variable `schema` are representing the current JSON value, not a JSON schema. I think for clarity, we might want to rename these to something more representative of the data, maybe `configJSON` or `config` like abvoe.

jafeltra (commented):
This is looking really good! I had a couple really small comments. Let me know what you think!

electron/electron.js
SenecaAnderson: I have renamed all of the schema variables to configJSON.

react-app/src/components/ConfigEditor.js
SenecaAnderson: I have renamed this variable

electron/electron.js
SenecaAnderson: I have made .json the default extension

jafeltra (approved):
Thanks for making those changes! This looks good to me!

PR #20, opened 2021-08-02
Load framework config

SenecaAnderson:
### Summary
The extraction process has been modified to accommodate the change to the config parameter of the mcodeApp function. Backend functions have been added to retrieve the config file schema.

### New Behavior
The app behavior hasn't visibly changed.

### Code Changes
- During extraction, extraction.js calls the getConfig function from the framework. This function takes the file path and returns the JSON object in that file. This object is passed to mcodeApp instead of the file path.
- schemaFormUtils, preload.js, and electron.js now have a chain of functions to retrieve the config file schema from the framework.
- Updated the mcode-extraction-framework dependency to the latest version of the develop branch

### Testing Guidance
Start the app with npm start. Run extraction. It will not throw an error related to the data type of the config file.
In order to see the schema returned by the new functions, add this import statement to ConfigForm.js:
`import { getConfigSchema } from './SchemaFormUtils';`
Add this code inside the ConfigForm function:
`getConfigSchema().then((schema) => {
    console.log(schema);
  });`

Go to the configuration editor. Click on "Create New". The config schema will log in the console.

### DIFF
--- a/electron/electron.js
+++ b/electron/electron.js
@@ -2,7 +2,7 @@ const { app, BrowserWindow, dialog, ipcMain } = require('electron');
 const squirrel = require('electron-squirrel-startup');
 const fs = require('fs');
 const path = require('path');
-const { logger } = require('mcode-extraction-framework');
+const { configSchema, logger } = require('mcode-extraction-framework');
 const runExtraction = require('./extraction');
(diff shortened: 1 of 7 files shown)


mgramigna (approved):
Looks good

PR #21, opened 2021-08-06
Render config form

SenecaAnderson:
### Summary
ConfigForm now renders a react-jsonschema-form Form using the schema from the framework. uiSchema, widgets, and fields objects have been created and are used by the Form.

### New Behavior
ConfigForm renders a Form.

### Code Changes
- Added configSchema state variable to ConfigEditor. This variable has to be stored in the ConfigEditors so that its value can be set when either "Create New" or "Load" is clicked; storing the schema in ConfigForm requires setting its value in the main body of the ConfigForm function, triggering an infinite render loop.
- Added <Form /> to JSX of ConfigForm, and imported associated values from schemaFormUtils
- Uncapitalized first letter of schemaFormUtils file name
- Added uiSchema, widgets, and fields objects to schemaFormUtils
- Ensured that CSS classes could be set using the uiSchema

### Testing Guidance
Start the app with npm start. Click on "Configuration Editor". Click on either "Create New" or "Load". The form generated by react-jsonschema-form will appear.


dmendelowitz (approved):
Looks good to me!

PR #22, opened 2021-08-09
Render extractor array

SenecaAnderson:
### Summary
The config form now renders an accordion for the extractor array. Each extractor has its own accordion item. There is a dropdown to add new extractors. There is a button to sort the selected extractors by type; this button is disabled until there is at least one extractor added

### New Behavior
The user can select extractors from a dropdown to add them to the array. The user can sort the extractors by type.

### Code Changes
- Added ExtractorArray and Extractor components to schemaFormUtils.js
- Added CSS styles to Pages.scss
- Placed container around <Form /> in ConfigEditor so that there is a margin around the form, instead of inputs starting at stopping at the very edge of the window

### Testing Guidance
Start the app with npm start. Click on "Configuration File Editor". Click on either "Create New" or "Load File". In the config form, scroll down to the extractors section. Try to hover over or click the "Sort by Type" button; it's disabled, so nothing will happen. Click on "Add new extractor" and select an extractor. The extractor will appear below the buttons in the accordion; its name will disappear from the "Add new extractor" list; and the "Sort by Type" button will no longer be disabled. Hover over and click on the sort button; nothing will happen, since there's only one extractor. Add as many extractors as you want, in any order. Click on the sort button to sort by type. Add more extractors and click the button again; the extractors will sort by type again.

### DIFF
--- a/react-app/src/components/schemaFormUtils.js
+++ b/react-app/src/components/schemaFormUtils.js
@@ -1,3 +1,6 @@
+import React, { useState } from 'react';
+import { Accordion, FloatingLabel, Form } from 'react-bootstrap';
+
 function getConfigSchema() {
   return window.api.getConfigSchema();
 }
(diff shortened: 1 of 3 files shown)


react-app/src/components/schemaFormUtils.js
```diff
@@ -67,8 +70,130 @@ const uiSchema = {
...
+    'CSVCancerDiseaseStatusExtractor',
+    'CSVCancerRelatedMedicationExtractor',
+    'CSVClinicalTrialInformationExtractor',
+    'CSVConditionExtractor',
+    'CSVObservationExtractor',
+    'CSVPatientExtractor',
+    'CSVProcedureExtractor',
+    'CSVStagingExtractor',
+    'CSVTreatmentPlanChangeExtractor',
+    'Extractor',
```
dmendelowitz, line 99: Since the Extraction UI only works with CSV Extraction, you could probably cut this list down to just any CSV[blank]Extractors. At the very least, you can definitely remove "Extractor" because that's just a base class that all other extractors extend, it doesn't actually do anything.

react-app/src/components/schemaFormUtils.js
```diff
@@ -67,8 +70,130 @@ const uiSchema = {
...
+      type,
+      constructorArgs: {
+        filePath: 'No File Chosen',
+      },
+    };
+  }
+
+  function addExtractor(eventKey) {
+    // Remove selected type from dropdown
+    const newTypes = types.filter((type) => type !== eventKey);
```
dmendelowitz, line 134: It's not something I had ever thought about before, but is there a scenario where someone would want to add 2 extractors of the same type? It seems totally possible, but I'm not sure.

dmendelowitz: For example: Someone has 2 CSVs of Patient data, one needs to be masked and the other doesn't, they'd have to add 2 Patient extractors with different settings, right?

dmendelowitz (commented):
Functionally, this works great! I just had 2 comments about what behavior we actually want to support. I wasn't totally sure about these though, so while I don't think this needs a second reviewer, I would appreciate a second opinion on my comments (@Dtphelan1 @mgramigna)

react-app/src/components/schemaFormUtils.js
Dtphelan1: I actually think you're right here Dylan. At first blush, I thought the second extractor would just overwrite the data of the first; but if the data stored in the two CSV's are totally different this would actually work. Very interesting. 

With that said, maybe we allow for the insertion of multiple extractors of the same type.

react-app/src/components/schemaFormUtils.js
SenecaAnderson: I have removed the extra extractors.

react-app/src/components/schemaFormUtils.js
SenecaAnderson: I changed this so that types aren't removed from the array when they're added. Additionally, types is no longer a state variable.

dmendelowitz (approved):

PR #24, opened 2021-08-11
Custom string input

SenecaAnderson:
### Summary
The initial task was to create a custom string input to be rendered by the react-jsonschema-form Form for string fields. However, this turned out to be unnecessary. The width issue was fixed with CSS classes instead.

### New Behavior
String fields are now limited to a max-width of 300px.

### Code Changes
- Added input-width-limit to Page.scss
- Modified CSS classes in uiSchema - added input-width-limit where appropriate
- Modified CSS classes in uiSchema - removed page-text from all fields, since it was only added for testing purposes in a previous task and is no longer necessary.
- Export statement in schemaFormUtils was once again fixed

### Testing Guidance
Start the app with npm start. Click on "Configuration Editor". Click on "Create New" or "Load". Many fields in the form will now have a maximum width of 300px. Fields that will be modified in other tasks (such as file inputs and fields in the extractor array) do not have a limited width.

### DIFF
--- a/react-app/src/components/schemaFormUtils.js
+++ b/react-app/src/components/schemaFormUtils.js
@@ -211,26 +211,12 @@ function FileWidget(props) {
   );
 }
 
-function StringInput(props) {
-  console.log(props);
-  return (
-    <Form.Group className="mb-3" controlId="props.name">
-      <Form.Label>
-        {props.schema.title}
-        {props.required && '*'}
-      </Form.Label>
-      <Form.Control type="text" />
-    </Form.Group>
-  );
-}
-
 const widgets = {
   FileWidget,
 };
 
 const fields = {
   ArrayField: ExtractorArray,
-  StringField: StringInput,
 };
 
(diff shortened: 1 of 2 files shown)


react-app/src/components/schemaFormUtils.js
```diff
@@ -1,5 +1,5 @@
 import React, { useState } from 'react';
-import { Accordion, Button, Dropdown } from 'react-bootstrap';
+import { Accordion, Button, Dropdown, Form } from 'react-bootstrap';
```
mgramigna, line 2: I'm seeing a lint warning saying we don't use `Form` anymore in this file. Can it be deleletd?

SenecaAnderson:
I have removed Form from the imports.

jafeltra (approved):
This looks good! I don't see the warning anymore and I think the narrower input fields look good, and will especially look nice if we have the two columns of input fields eventually.

PR #25, opened 2021-08-24
Render extractors

SenecaAnderson:
### Summary
This branch adds an input form to each extractor in the extractor array, as well as the ability to delete arguments.

### New Behavior
When new extractors are added to the extractor array and the accordion header of an extractor is clicked, a form appears. The form has inputs for the label and constructor args. If filePath and url are the only constructor arg options for that type, then only those will appear, with a button to toggle between them (default is filePath). If there are other constructor arg options, a dropdown will appear with those options. When an argument is selected from the dropdown, that argument will appear with the appropriate form input. Each has a trashcan icon that can be clicked to delete the constructor arg.

### Code Changes
- Moved ExtractorArray and Extractor functional components to their own files
- Added state variables to Extractor in order to manage the constructor args
- Modified JSX in Extractor
- Modified ExtractorArray to accommodate changes to Extractor
- Changes to CSS styles in Pages.scss

### Testing Guidance
Start the app with npm start. Click on "Configuration Editor". Add extractors using the dropdown. Fill out the forms, then sort by type; the data in the forms will stay with the correct extractor. Extractors with additional constructor arg options (check args state variable to see what extractors are valid for each arg) will have a dropdown that allows the user to add those args. The dropdown shows disabled styling once there are no more args to add. 
Click on "Switch to URL" and "Switch to CSV File" to toggle between the filePath and url options.
Click on the trashcan icon in the accordion headers to delete extractors. Click on the trashcan icon next to extra constructor args to delete those args.


react-app/src/components/Extractor.js
```diff
@@ -0,0 +1,303 @@
...
+      },
+    ];
+  }
+  const [args, setArgs] = useState(getArgValues());
+  const [argsJSX, setArgsJSX] = useState(
+    getArgValues()
+      .filter((arg) => arg.included === true)
+      .map((arg, i) => (
+        <p key={arg.key}>
+          This is a placeholder for arg {i}: {arg.label}
```
jafeltra, line 75: Is there ever a value we use the placeholder for anymore? If this is just used as the default value, is there a simpler value we can use here?

react-app/src/components/Extractor.js
```diff
@@ -0,0 +1,303 @@
...
+            return (
+              <Form.Group className="mb-3" controlId={arg.key} key={arg.key}>
+                <Form.Label>{arg.label}</Form.Label>
+                <div className="label-and-icon-container">
+                  <Form.Control
+                    type="text"
+                    value={arg[arg.key]}
+                    onChange={(e) => {
+                      const newArgs = [...tempArgs];
+                      newArgs.find((temp) => arg.key === temp.key)[arg.key] = e.target.value;
```
jafeltra, line 102: I am curious what others think of this pattern, but there are a few places where you do a `find` and the directly manipulate the resulting object. While it doesn't make a big difference in the functionality of the code itself, I think it might make it easier to read this code in two lines like

```
const currentArg = newArgs.find((temp) => arg.key === temp.key);
currentArg[arg.key] = e.target.value;
```

There are a number of places with this pattern, so if we update this one, we should also update the others.

react-app/src/components/Extractor.js
```diff
@@ -0,0 +1,303 @@
...
+                <div className="label-and-icon-container">
+                  <Form.Control
+                    type="text"
+                    value={arg[arg.key]}
+                    onChange={(e) => {
+                      const newArgs = [...tempArgs];
+                      newArgs.find((temp) => arg.key === temp.key)[arg.key] = e.target.value
+                        .split(',')
+                        .map((word) => word.trim());
+                      updateArgData(newArgs, false);
```
jafeltra, line 209: If possible, I think it would be good to separate the manipulation you do on the string from what is actually displayed in the UI. With this how it is now, the input field swallows the spaces when you type them, which is a little jarring. Also, if you try to hit enter while typing in the middle of the string, the cursor is moved to the end, which isn't expected. I think it would be good if we only did the `split` and `trim` when we use the formData to actually output the configuration file, and not update what is actually input into the box.

I'm not sure if we actually want to or can fix this now, but it might be good to be aware of when we hook up with submit button and access all the data from the form.

react-app/src/components/Extractor.js
```diff
@@ -0,0 +1,303 @@
...
+                    className="mouse-pointer"
+                  />
+                </div>
+              </Form.Group>
+            );
+          default:
+            return (
+              <p key={arg.key}>
+                This is a placeholder for unidentified argument {i}: {arg.key}
+              </p>
```
jafeltra, line 228: I know there aren't any cases right now where we hit this, but I think it would be good to just have this be a blank input box. If we do that, we are no longer limiting what can be created in the UI compared to what can be made in a text editor. An input box may not be particularly useful, but it will at least let you do the same thing you can do in an editor (unless you're trying to input a number or a boolean, but we can decide if that's worth handling later).

jafeltra (commented):
This is looking really good. I had a few comments and questions inline, as well as a few overall interactions below:

[JA1] When the URL field is toggled to, I think the label should be "CSV File URL", which more closely lines up with what the label is for the file path option.
[JA2] I remember at one point we talked about how the accordion starting closed was helpful when trying to trigger updates at the right time, but is that still true? If it isn't, I think it might be nice to have the new extractor you add start of open. When I had more than 4 extractors in the list, I found myself getting confused because I had an extractor open that was not the one I had just added, and my new one was really small and closed at the bottom of the list.
[JA3] When you have added all the constructor args that apply to a certain extractor, the dropdown just does nothing. I think it could be nice to have a greyed out option that says "All arguments added" or "No available arguments".
[JA4] If I add a constructor argument (for example, site id), enter a value, and then delete the constructor argument, when I re-add the the constructor argument, the value is still populated. I'm curious if this was intentional. I was expecting it to be cleared out, but didn't know if I was missing a case where it could help to remember the last value typed.

Of these, I think JA1 is the only one that should definitely be addressed as part of this PR. The others we can discuss if we agree and if they should be a part of this PR or separate tasks.

react-app/src/components/Extractor.js
```diff
@@ -0,0 +1,303 @@
...
+        key: 'cancerType',
+        validExtractors: ['CSVCancerDiseaseStatusExtractor'],
+      },
+      {
+        mask: props.formData.constructorArgs.mask ? props.formData.constructorArgs.mask : '',
+        label: 'Masked Fields',
+        type: 'dropdown',
+        included: false,
+        key: 'mask',
+        options: [`gender`, `mrn`, `name`, `address`, `birthDate`, `language`, `ethnicity`, `birthsex`, `race`],
```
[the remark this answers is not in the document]
SenecaAnderson: I have changed these to regular strings.

react-app/src/components/Extractor.js
SenecaAnderson: This is the JSX that is replaced when the user clicks on the accordion header, so it's never visible; the placeholder is used because the real form requires functions that don't exist yet, and those functions require this state variable to already exist. I have replaced the <p> and text with an empty div.

react-app/src/components/Extractor.js
SenecaAnderson: I have replaced this code.

react-app/src/components/Extractor.js
SenecaAnderson: I can leave it as a regular string input for now. Then I can add the code to convert the string -> array while working on STEAM-723.
I have removed the calls to split and trim, so now it just saves the string.

react-app/src/components/Extractor.js
SenecaAnderson: So I should add a new string input?

SenecaAnderson:
[JA1] I have changed the label.

[JA2] Clicking on the accordion header is what causes the JSX to update so that the form displays instead of the placeholder. I will look into using the activeKey prop to manage which accordion is open, and if this can be used without causing the placeholder to appear. 
Additionally, I could add new extractors to the start of the array, instead of the end. Would that make it less confusing, even if it turns out that I can't use the activeKey to manage which accordion is open?

[JA3] I have modified this. When all the arguments are added, the dropdown switches to a <Button/> with disabled styling. The Button says "All arguments added". 

[JA4] This was intentional; if the user deletes the argument, then changes their mind, then they can add it again and get the old value back. While the deleted value is stored in the args array, it's not saved to the formData object; the field disappears from the formData object entirely. I can delete the value when the arg is deleted if you want.

react-app/src/components/Extractor.js
SenecaAnderson: Should I make a function to get the value for currentArg? This code repeats a lot, and that might simplify things.

react-app/src/components/Extractor.js
jafeltra: Yeah I think it makes sense to have this be a string input!

jafeltra:
Thanks for making the updates! I'll take a look at them!

[JA2] I think it's worthwhile to investigate using the activeKey to manage which is open. However, if you can't do that, I think keeping them added at the end is okay!

[JA4] Ah that's good that it isn't saved int he formData object. I think it makes sense to keep the value there for now in that case, so no need to update anything for this one!

react-app/src/components/Extractor.js
SenecaAnderson: What variable should I connect it to? The function to update the arg values doesn't exist yet, and if I set the formData object for the extractor equal to a string, that could make it incompatible with the react-jsonschema-form schema.

react-app/src/components/Extractor.js
```diff
@@ -0,0 +1,312 @@
...
+        validExtractors: ['CSVPatientExtractor'],
+      },
+    ];
+  }
+  const [args, setArgs] = useState(getArgValues());
+  const [argsJSX, setArgsJSX] = useState(
+    getArgValues()
+      .filter((arg) => arg.included === true)
+      .map((arg) => <div key={arg.key} />),
+  );
```
jafeltra, line 74: I'm actually now thinking we don't need to manage this state variable at all. I think since `argsJSX` uses args to render the various inputs, we can actually just call `getArgsJSX` from within the render function. If we do that, then I don't think we need to save and keep track of the resulting value because we can just directly render it, and re-render it, when the data updates.

react-app/src/components/Extractor.js
```diff
@@ -0,0 +1,312 @@
...
+  }
+
+  function updateArgs(inputArgs, isAccordion = true) {
+    let tempArgs;
+    if (isAccordion === true) {
+      tempArgs = [...args];
+    } else {
+      tempArgs = [...inputArgs];
+    }
+    setArgsJSX(getArgsJSX(tempArgs, updateArgs));
```
jafeltra, line 244: If we get rid of the `setArgsJSX` state variable, I think we can just call` getArgsJSX` directly from within the render function, where `argsJSX` is now.

I also think that if we move this out of the `updateArgs` function, we can just use `updateArgs` directly within the `getArgsJSX` function and remove it from the parameters of the function. I think this will simplify the code and will keep us from passing around a reference to a function that we have direct access to.

react-app/src/components/Extractor.js
```diff
@@ -0,0 +1,312 @@
...
+  }
+
+  function addArg(eventKey) {
+    const tempArgs = [...args];
+    const currentArg = tempArgs.find((arg) => arg.key === eventKey);
+    currentArg.included = true;
+    updateArgs(tempArgs, false);
+  }
+
+  function getHasArgs() {
```
jafeltra, line 264: This function looks like it's just checking if an extractor has argument to add. Can we simplify the name and just call it `hasArgs`?

jafeltra (commented):
I thought of a few other things that I think will help to simplify the code. Let me know what you think!

react-app/src/components/Extractor.js
jafeltra: Oh hm that's a good point. We don't want to build in something that won't validate and there's no way to change it in order to validate it. Maybe this placeholder is actually the way to go for now. There isn't any current argument that reaches this case, right?

SenecaAnderson: If we can get rid of argsJSX, then we won't need the placeholder at all anymore

react-app/src/components/Extractor.js
SenecaAnderson: Yeah. It would be:
`const currentArg = getCurrentArg(arg.key);`

react-app/src/components/Extractor.js
SenecaAnderson: There's no argument that reaches this case

react-app/src/components/Extractor.js
SenecaAnderson: I have removed the argsJSX variable.

react-app/src/components/Extractor.js
SenecaAnderson: I have removed the setArgsJSX call and moved updateArgs above getArgsJSX.

react-app/src/components/Extractor.js
SenecaAnderson: I have changed this function's name to hasArgs.

react-app/src/components/Extractor.js
SenecaAnderson: I have removed argsJSX, so this placeholder no longer exists.

SenecaAnderson:
The update for managing which accordion is active is currently in progress.

PR #26, opened 2021-08-24
Fixed getFile file types bug

SenecaAnderson:
### Summary
The getFile process, which previously only allowed the user to select JSON files, now accepts an array of file extensions to determine which file types are allowed. The default is to allow any file type.

### New Behavior
None. The changes aren't visible in the front end.

### Code Changes
- window.api.getFile now has a parameter called 'extensions' with a default value of an empty array
- the get-file invoke and handler functions now receive and pass on the extensions array
- the handler uses the extensions array when passing options to the electron dialog
- calls to window.api.getFile have been modified to use the new parameter

### Testing Guidance
Use npm start to start the app. Go to the extraction and configuration processes and use the file pickers. They will only allow you to select the appropriate file type. At the moment, the only file picker that uses CSVs (as opposed to JSON files) is the patient MRNs picker on the configuration editor form.

### DIFF
--- a/electron/electron.js
+++ b/electron/electron.js
@@ -65,9 +65,9 @@ app.on('activate', () => {
 
 ipcMain.handle('get-config-schema', async () => configSchema);
 
-ipcMain.handle('get-file', async () =>
+ipcMain.handle('get-file', async (event, extensions) =>
   dialog.showOpenDialog(mainWindow, {
-    filters: [{ name: 'JSON', extensions: ['json'] }],
+    filters: [{ name: 'extensions', extensions }],
     properties: ['openFile'],
   }),
 );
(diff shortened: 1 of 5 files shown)


electron/electron.js
```diff
@@ -65,9 +65,9 @@ app.on('activate', () => {
 
 ipcMain.handle('get-config-schema', async () => configSchema);
 
-ipcMain.handle('get-file', async () =>
+ipcMain.handle('get-file', async (event, extensions) =>
   dialog.showOpenDialog(mainWindow, {
-    filters: [{ name: 'JSON', extensions: ['json'] }],
+    filters: [{ name: 'extensions', extensions }],
```
jafeltra, line 70: Do you know what the `name` field is for? If it's a name relevant to the dialog itself, I'm not sure if `'extensions'` is an informative name. Maybe we don't need to add it at all?

jafeltra (commented):
It's awesome to be able to select a CSV in the "Patient ID CSV File" field! I just had one small question inline.

electron/electron.js
SenecaAnderson: filters is an array of FileFilter objects. Each of the FileFilters has a name. I'm not sure what the name is actually for; I think so you can differentiate between the different objects? But the name isn't displayed anywhere, and deleting it makes the file filter stop working.

PR #27, opened 2021-08-27
Add navbar & create new app name

SenecaAnderson:
### Summary
This PR adds a Navbar to most pages of the app ~(excluding the processing page and the results page)~. Additionally, the app has been named ~UniFHIR~ mCODE Extractor.

### New Behavior
A Navbar appears on ~most pages~ all pages of the app. When the window is wide, the links (home, extractor, and config editor) appear in the bar; when the window is narrow, the Navbar collapses and has a hamburger menu. When you click on the hamburger, a dropdown with the links in it appears. Clicking on the app name redirects to the home page. Additionally, the styling of the CSS class page-title has been changed to make things look more balanced now that there's a Navbar above the titles.
~The results page doesn't have a Navbar. This is because including the Navbar causes major issues with the display of the logger messages, and it looks super weird with the sidebar header, which is the same color and has bigger text. Some aspects of how the results page looks may need to be reevaluated in order to accommodate a Navbar.~ This is done now!

### Code Changes
- Updated Bootstrap and React-Bootstrap to latest versions
- Added Navbar component
- Changes to CSS styles in Pages.scss
- Changed app name at start of README

### Testing Guidance
Start the app with npm start. The Navbar appears on ~the aforementioned pages~ all pages. To show the expanded version, make the page wider; to make it collapse, make the page smaller. Clicking on the links in the Navbar will send you to the appropriate pages. Clicking on ~"UniFHIR"~ "mCODE Extractor" will bring you to the homepage.

**Edited by @jafeltra after commits https://github.com/mcode/mcode-extraction-ui/commit/ed36a4f21cd2ec8b2d8de288214905eaa045e968 and https://github.com/mcode/mcode-extraction-ui/commit/7ea12985699ee8dd28a1dd2faf5d21b6e31bad63.**


jafeltra:
I looked at this PR and I think it might be worth exploring how to appropriately get the NavBar to display on the Results page. I find it jarring to have a NavBar disappear on the page. I've been looking into it, and it isn't straightforward, but I think it might be better to fix it as part of this PR, rather than merge in something we will want to change immediately.

I also think it would be good to discuss the name as a team before merging in this PR. I love the UniFHIR pun, but we should decide if this conveys what the tool does and if we want to incorporate STEAM or Extraction in the name at all. This can be a next week discussion, but I wanted to post what I had thought about while reviewing this.

jafeltra:
After discussing with the team, I've renamed the app to "mCODE Extractor". That change is in https://github.com/mcode/mcode-extraction-ui/commit/7ea12985699ee8dd28a1dd2faf5d21b6e31bad63).

I've also updated the Nav Bar so that it appears on all the pages. As part of that update, I had to update the layout of a few pages. This led me down the path of updating where the navigation buttons (back and submit/save) were on most pages. That led to a few positioning issues in the app, but I think I've handled all of them now. The layout/navigation is still by no means perfect, but hopefully this is an improvement. There are still tasks in the backlog to improve it further. Also, the initial home page has some weird spacing on it, but I didn't spend time adjusting it since we have a task a the very top of the backlog to get rid of that home page and replace it with the "Extract New" page. All these changes are made in https://github.com/mcode/mcode-extraction-ui/commit/ed36a4f21cd2ec8b2d8de288214905eaa045e968.

If you'd like to compare these changes to what the branch was before I started changing things, check out commit https://github.com/mcode/mcode-extraction-ui/commit/4ce2bd27038fc5f3a6e1bd35359546742ffcf684.

This is ready to be reviewed by someone else now!

Dtphelan1:
DP1 – I don't think the hamburger menu actually opens/works right now. Feel free to confirm that it works on your machine and show me that I'm just a jabroni, but in the event it doesn't work on yours either that should be fixed.

Dtphelan1:
DP2 - any reason we're using the `container` class on the contents of the navbar and not throughout the rest of the app? I think if we should go for an all or nothing approach with the gutters. With that in mind, I'd recommend removing the `container` class for now since I don't know how adding gutters to the results page will mess with the look/feel.

jafeltra:
> DP1 – I don't think the hamburger menu actually opens/works right now. Feel free to confirm that it works on your machine and show me that I'm just a jabroni, but in the event it doesn't work on yours either that should be fixed.

Oops I definitely broke it. I'll fix it.

Dtphelan1:
DP3 - the content of the home page is missing a top padding. I think this is typically accomplished by wrapping the content in a `div.page-content` but now the wrapping div has no class. Could we add that class as appropriate?

jafeltra:
> DP2 - any reason we're using the `container` class on the contents of the navbar and not throughout the rest of the app? I think if we should go for an all or nothing approach with the gutters. With that in mind, I'd recommend removing the `container` class for now since I don't know how adding gutters to the results page will mess with the look/feel.
> 
> EDIT: Another approach (which we've started in a few places) is to use the Container component with the fluid property. See the ResultPage component for addl. clarification

Hm great question! I think that's something Seneca added and I just didn't touch. I was getting thrown off by the Container class that had the padding/margin (I don't remember which it was). I'll try to align them all!

Dtphelan1:
> > DP2 - any reason we're using the `container` class on the contents of the navbar and not throughout the rest of the app? I think if we should go for an all or nothing approach with the gutters. With that in mind, I'd recommend removing the `container` class for now since I don't know how adding gutters to the results page will mess with the look/feel.
> > EDIT: Another approach (which we've started in a few places) is to use the Container component with the fluid property. See the ResultPage component for addl. clarification
> 
> Hm great question! I think that's something Seneca added and I just didn't touch. I was getting thrown off by the Container class that had the padding/margin (I don't remember which it was). I'll try to align them all!

That's the one without the fluid property. Try adding that 😄

jafeltra:
> DP3 - the content of the home page is missing a top padding. I think this is typically accomplished by wrapping the content in a `div.page-content` but now the wrapping div has no class. Could we add that class as appropriate?

I can do that, but I also have removed that page entirely for the next task. Is that task intended to remove the component entirely?

Dtphelan1:
> > DP3 - the content of the home page is missing a top padding. I think this is typically accomplished by wrapping the content in a `div.page-content` but now the wrapping div has no class. Could we add that class as appropriate?
> 
> I can do that, but I also have removed that page entirely for the next task. Is that task intended to remove the component entirely?

lol if you've already started and nearly finished the _remove this page_ task, then nvm 😄

jafeltra:
DP1 - I addressed this by making the Navbar component use the "fixed" prop, which fixes the position to the top of the page. I used padding to move the remaining content down on the page. This way, when you click on the hamburger menu, the menu opens on top of the rest of the content. Let me know what you think.

DP2 - I added `fluid` to the container in the Navbar, which gets a little closer to lining up the gutters in the app. They're definitely still all over the place (which is something I hadn't even thought of until you pointed it out, so thank you), but I can look to align them all as part of this PR if we want.

DP3 - I'm not planning to make any changes to the home page because I removed it in another branch coming soon to a PR near you.

Let me know what you think!

Dtphelan1 (approved):
You've addressed all my comments. This looks like a great improvement to me!

jafeltra:
@Dtphelan1 thank you very much for your approval. Unfortunately, in an unideal display of teamwork, I added on a last little commit. It adds an href and the `collapseOnSelect` prop to properly close the Navbar when an option is selected from the hamburger menu (when the app is very narrow). If you wouldn't mind taking a brief look and making sure it still works as expected, I'd appreciate it.

Dtphelan1:
1) The collapseOnSelect looks great.
2) Where is the href motif coming from? Is that standard, and is that necessary to accomplish the collapseOnSelect? Just wondering.

jafeltra:
> 2\. Where is the href motif coming from? Is that standard, and is that necessary to accomplish the collapseOnSelect? Just wondering.

Asking the real tough questions. So I'm entirely sure why the `href` is needed. I think it is because we are using the `as` prop to render a react router `Link` and using the `to` prop to link appropriately. However, it looks like the `href` prop is what react bootstrap is expecting to be used as the navigation and it seems like it is tied in with the `collapseOnSelect` in the background. If you remove the `href`, the navigation works correctly in the app, but the menu doesn't close. If you remove the `as` and `to` props and keep the `href`, the navigation doesn't work, but the menu closes. I have a feeling we aren't using the `as` prop correctly, but I hadn't found another way to navigate using the NavBar with react router.

I really don't have any motivation for the value I put in each `href` since I don't think it will work correctly with react-router so they are kind of just placeholders. It's pretty hacky, so I can gladly remove it if you don't like the hack (which is very valid).

jafeltra:
34e6c07 should address the comments we talked about at stand up this morning!

Dtphelan1 (approved):
Looks great! Unideal post-approval code changes re-approved!

PR #28, opened 2021-09-10
Remove Home page and replace with Extract New page

jafeltra:
This PR removes the `Home` page and replaces it with the `Extract New` page.

All the paths that directed to `/extract` have been replaced with `/` since that is the new home page. If we think that in the future we might want a different landing/home page or we like the descriptiveness of `/extract`, I can revert those changes and just redirect `/` to `/extract` for now. This was another approach I considered and didn't end up going with, but is very possible.


react-app/src/components/Extract.js
```diff
@@ -153,7 +152,7 @@ function Extract(props) {
             </Row>
           </Form>
           <div className="nav-button-container">
-            <LinkButton className="generic-button" size="lg" variant="outline-secondary" text="Cancel" path="/" />
+            <div></div>
```
Dtphelan1, line 155: DP1 - Do we need this `div`?

Dtphelan1:
DP2 - Can we change the text on the "Back" button below to something more indicative? Do we even want a back button here anymore? I'd argue we might not, right – since the Config Editor is on it's own page now?

image

jafeltra:
DP1 and DP2 are addressed in 78a1c2f

Dtphelan1:
> What is but one more pixel?  
Everything

Dtphelan1 (approved):

PR #29, opened 2021-09-17
Configuration Form is now saved to disk 

julianxcarter:
This PR updates the 'Submit' button within the Configuration Editor to save the formData from the form component into a json file, rather than a placeholder object. 

I ended up using the Submit button that's provided as part of the `Form` component. This is because the only way to get the updated `formData` prop is by using the submit functionality built in to that component. I also moved the action bar containing the `Back` button from the `ConfigForm` to the `ConfigEditor` page because I felt that was a more intuitive location for it, but if others feel otherwise I can move it back.

### DIFF
--- a/react-app/src/components/ConfigEditor.js
+++ b/react-app/src/components/ConfigEditor.js
@@ -24,13 +24,21 @@ function ConfigEditor() {
 
   function toggleForm() {
     getConfigSchema().then((schema) => {
+      // eslint-disable-next-line no-param-reassign
+      delete schema.description;
       setConfigSchema(schema);
     });
     setShowForm(!showForm);
   }
 
+  function closeForm() {
+    setShowForm(false);
+  }
+
   function loadFile() {
     getConfigSchema().then((schema) => {
+      // eslint-disable-next-line no-param-reassign
+      delete schema.description;
       setConfigSchema(schema);
     });
     window.api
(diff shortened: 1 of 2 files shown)


dmendelowitz:
When I go through and edit a config and save the first time, this works great, but then if I go and try and save a second config, I get this error:
error
I think it might have something to do with the form getting cleared out after you hit save? Because all the text fields get cleared out after I hit save the first time, but the extractor forms stay, so maybe they just stay graphically but are actually cleared behind the scenes leading to the forEach undefined thing? It seemed kind of strange to me that the form gets cleared out after you save

julianxcarter:
> I think it might have something to do with the form getting cleared out after you hit save? Because all the text fields get cleared out after I hit save the first time, but the extractor forms stay, so maybe they just stay graphically but are actually cleared behind the scenes leading to the forEach undefined thing? It seemed kind of strange to me that the form gets cleared out after you save

This was correct! By default a form component will clear all values after submission, but because we're using custom fields for the MRN file picker and extractor array the UI was not being updated to reflect this. Hence the error you were seeing. The most recent commit causes the data you enter into the form to persist between saves.

dmendelowitz (approved):
With that fixed, I think this looks good! It makes more sense to not have the form clear out after saving anyways so I think that was a good change in general

PR #30, opened 2021-09-24
Add validation

Dtphelan1:
This task took a look at managing validation outside of the default mechanism and adds validation back to our config editor UI. 

## The Evaluation

Given the fact that this code won't be remotely hosted and is designed to be run locally, we've decided that adding `unsafe-eval` to the HTML's content security policy doesn't introduce undue risk to our users. Therefore, for usability's sake, we will move forward with the built-in validation support.

## Additional comments

In addition to flipping the validation switch this task ballooned into an investigation of existing issues with validation, resulting a few minor fixes and a bunch of future JIRA tickets. 

The fixes include: 
- Using the `window.maximize` function to full-screen the application on launch (this was going to be a subtask of STEAM-702, and the description has now been updated accordingly).
- Adding AJV so we can manually load the JSON Schema Draft 6 metaschema into our form validation.
- Minor linting of the configForm file.
- Fixing an erroneous override of the default ArrayField component with the ExtractorArray component. Instead, we add a new kind of Field to our `fields` object, and declare the use of that field in our uiSchema. This fixes previous issues with selecting Option 2 for our `notificationInfo.to` field not working. 
- (After the above was done) Adds CSS to properly style a few Array-item-action buttons, including moving an array element up, moving it down, removing it, and adding it.
- Removes CSS overriding the default bootstrap Row styling. I couldn't find where this was needed, but in the event that we do find the culprit Row in this review we can use the [default bootstrap utility classes](https://hackerthemes.com/bootstrap-cheatsheet/) to override the styling as needed. 

Additional tasks created include: 
- 752:    Evaluate (and act on) differences between JSON Draft Schema 6 to Schema 7 
- 753:    ExtractorArray Field doesn't prepopulate with formData
- 754:    ExtractorArray component should not manage data and view separately
- 755:    Explore and solve use of the `type` keyword in our schema
- 756:    FileWidget needs better affordances when failing validation.
- 757:    Basic housekeeping on the mCODE-Extraction-UI repo

## Review Goals

A review of this PR should:
 - Make sure validation _kinda_ works;
 - Work on quick-fixes for any small validation issues; 
 - Create tasks for larger validation woes;
 - Review the tasks mentioned above for correctness and legibility;
 - Confirm that small-fixes mentioned above work as expected.

### DIFF
--- a/electron/electron.js
+++ b/electron/electron.js
@@ -19,12 +19,11 @@ let mainWindow;
 const createWindow = () => {
   // Create the browser window.
   mainWindow = new BrowserWindow({
-    width: 800,
-    height: 600,
     webPreferences: {
       preload: path.join(__dirname, './preload.js'),
     },
   });
+  mainWindow.maximize();
 
   // load build assets if production, localhost URL otherwise
   const mainUrl = app.isPackaged
--- a/react-app/package-lock.json
+++ b/react-app/package-lock.json
@@ -14,6 +14,7 @@
         "@testing-library/user-event": "^12.1.10",
         "@types/react-router-bootstrap": "^0.24.5",
         "@types/react-router-dom": "^5.1.7",
+        "ajv": "^8.6.3",
         "bootstrap": "^5.1.0",
         "fhirpath": "^2.9.1",
         "lodash": "^4.17.21",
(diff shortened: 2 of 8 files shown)


dmendelowitz (commented):
I think this looks pretty good generally. All the small fixes look great and validation works in all the simple fields, but there's some weird behavior in some cases. Here's what I noticed (some of these might be covered by the other tasks you made, ignore them if they are):
-  There are 2 "should be string" messages when I start the config editor, presumably because there's no value in the fields to start with
- It seems the custom formats for "from" and "to" (emails separated by commas and emails with names) aren't recognized by react json schema. Right now for "to" the array (option 2) will properly enforce email format while option 1 doesn't and for "from" it just accepts any string
- I don't know the exact specification of the "hostname" format, but I couldn't make it give me an error for anything I entered in the "host" field
- When Patient ID CSV file path is left unchosen, it actually outputs "No File Chosen" as the value to the JSON file, so there's no validation error when you don't choose a file for that field

Dtphelan1:
1. `should be string` actually refers to the fact that the property `extractors[].type` is not a string, but it expects it to be. Given that, we should only expect one warning. 
2. The custom formats definitely need to be added. I could be convinced to look into that as a part of this task, but my initial thought is to make that a different task. LMK if you have a strong feeling one way or another. 
3. I'm also not terribly familiar with that specification... I will look into that one.
4. I'll look into if that's a problem with how we model that data in our Custom Widget.

dmendelowitz:
to
The errors for "to" weirdly appear twice for me. Also somewhat confusing that it says "should be array" even when its not in Option 2. Maybe because its validating both sides of the "anyOf" and notices my input doesn't match either? I know custom formats have their own task in the backlog so maybe we don't need to fix these now unless its an easy fix. Other than that I think this looks really good!

Dtphelan1:
@dmendelowitz I've created a task for the above comment. Let me know if there is anything else I should address for an approval!

dmendelowitz:
With the new style changes, I don't actually see the label "To" above the related text field
no_to

Dtphelan1:
@dmendelowitz This seems to be the way `RJSF/bootstrap-4` chooses to handle AnyOf/OneOf properties nested on an object. I'm going to look at improving this in the other task I'm assigned, STEAM-761.

dmendelowitz (approved):

PR #31, opened 2021-10-05
Option picker for Masking Patient values

julianxcarter:
This PR introduces a picker for the `mask` constructorArg within the patientExtractor config section of the Config Editor. Though they're not currently functional, the mask values [that will soon be added to the MEF](https://github.com/mcode/mcode-extraction-framework/pull/155) are included options as well.


Dtphelan1:
DP1 - We could probably update the MEF version now 😄

Dtphelan1:
DP2 - Needs a rebase

Dtphelan1:
DP3 - Small thing, but I noticed that if you click on the label of a checkbox (e.g. the text-label `telecom` next to the checkbox) it alternates the checked-value for the wrong field. After doing some digging, I learned that providing an `id` prop to the Form.Check component will solve this issue (see https://github.com/react-bootstrap/react-bootstrap/issues/4056#issuecomment-755458634). 

tl;dr let's add an `id` prop to the Form.Check component that's identical with the `key` prop.

PR #38, opened 2021-10-18
Error filtering in the configuration editor

julianxcarter:
Addresses some of the oddities we've encountered with the `comma-separated-emails` validation in the Configuration Editor. There were two separate issues going on here; the first being that two seemingly unrelated errors were being returned by the validator when validation for this field failed. That issue is addressed in this PR. The second issue is that validation errors for that field are rendered by RJSF twice, and [an issue for that has been filed on the RJSF repo](https://github.com/rjsf-team/react-jsonschema-form/issues/2584).

To test, start the Extraction UI and navigate to the Configuration Editor. Fill out the Notification Info section with the `comma separated emails` option for the To field and an invalid list of email addresses. Note that on master, the following errors should be included that are not rendered on this branch:
* `should be array`
* `should match exactly one schema in oneOf`

### DIFF
--- a/react-app/src/components/ConfigForm.js
+++ b/react-app/src/components/ConfigForm.js
@@ -1,6 +1,7 @@
 import React, { useState } from 'react';
 import { Alert, Button } from 'react-bootstrap';
 import Form from '@rjsf/bootstrap-4';
+import _ from 'lodash';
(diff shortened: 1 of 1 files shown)


Dtphelan1 (approved):
This looks great to me. Much more helpful error messaging.

PR #41, opened 2021-10-28
Fix validation in the Configuration Editor

julianxcarter:
This PR addresses the errors we've been having with validation on second runs in the Configuration Editor. This problem was due to some wonkiness in how RJSF validates elements that use the `oneOf` keyword in their JSON schema. To address this, I've created a separate version of the config schema without the `oneOf` keyword and added that version to this repository. As a consequence, users of the UI will be limited to the `array` options in the NotificationInfo.to and Patient Masking sections of the configuration editor. As an added perk, this eliminates the dual error messages that were being rendered in the NotificationInfo section.

Additionally, I've done some slight reorganizing of the `react-app/src/components` directory. Changes include the following:
- Creating 4 subdirectories based on the page that each component is rendered in: `CommonComponents`, `ConfigEditor`, `ExtractNew`, and `Results`. Imports have been updated accordingly.
- Adding a `helpers` subdirectory to the Results and ConfigEditor directories. These contain `patientUtils.js` and `schemaFormUtils.js`, respectively.
- `ExtractorArray.js` has been renamed to `ExtractorAccordion.js`. The module export has been renamed as well.
- `ConfigForm.js` has been renamed to `EditorForm.js`.
- `Extractor.js` has been renamed to `ExtractorObject.js`

If we're not a fan of some or all of these organization choices, then I can roll them back and save them for the larger refactor task that we have!


jafeltra (approved):
This looks great! I feel so powerful saving a config file twice in a row 😎

PR #42, opened 2021-10-29
Use latest MEF and mask fields

jafeltra:
Very small PR that updates the MEF version and uses the new `genderAndSex` field in the UI to reflect the current MEF value.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4601,7 +4601,7 @@
     },
     "node_modules/mcode-extraction-framework": {
       "version": "1.1.0",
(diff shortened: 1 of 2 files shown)


julianxcarter (approved):
This looks good! Just needs a rebase and I think it'll be good to merge 👍🏽

jafeltra:
@julianxcarter I rebased... and then realized I forgot 50% of the task, which was removing `birthsex` from the options. So I did that.

julianxcarter (approved):
Well... we can all just collectively forget that I approved this when it only did 50% of the task 😅. This looks great!

PR #43, opened 2021-11-02
Electron config

Dtphelan1:
Definition of Done: 
    - [x] Updating the icon/logo that shows up in the dock   
    - [x] Updating the name that shows up in the menu bar/dock hover
        - Comment: The name _will_ update when the application is bundled, but it seems to refuse to do that before hand.
    - [x] Updating the title bar to something more than "React App"
    - [x] Open app in full screen width by default
        - Comment: This was done in a prior PR

Additionally, I added a Downloads folder bounce animation on the Mac OS when a saved file goes into the Downloads folder. This should work for both Config saving and patient-data saving.

Testing should ensure: 
- That all aspects of definition of done are met 
- That bundling the application with `npm run build` creates an application that has the appropriate Title and Logo 
- That saving config files still works 
- That saving patient data sitll works


electron/electron.js
```diff
@@ -15,16 +15,30 @@ if (squirrel) {
...
+    app.dock.downloadFinished(path.join(savePath));
+  }
+}
+
 // declare mainWindow ahead of time so that it can be accessed in .handle('get-file')
 let mainWindow;
+// Path to Icon file
+const iconPath = path.join(__dirname, 'static/icon.png');
+// Set the icon in the Mac OS Dock
+app.dock.setIcon(iconPath);
```
julianxcarter, line 30: It looks like even when the app is built using `npm run build`, the icon is only set once the app is running. The stock electron icon remains in use both within Finder and for a second once the app starts or closes.

https://user-images.githubusercontent.com/53443041/140146066-ec8203a3-0ca2-4876-8bef-819e19f2d0c9.mov

I remember us running into this issue when we first tried to change the icon and I'm not quite sure what the solution is 🥴. This may be something that occurs during distribution, and if that's the case then I think we should just save this for a separate task.

electron/electron.js
```diff
@@ -127,6 +142,7 @@ ipcMain.handle('save-output', async (event, savePath, outputBundles, saveLogs) =
     const outputFile = path.join(savePath, `mcode-extraction-patient-${data.index + 1}.json`);
     fs.writeFileSync(outputFile, JSON.stringify(data.bundle, null, 2), 'utf8');
   });
+  downloadBounceOnMac(savePath);
```
julianxcarter, line 145: For some reason this bounce animation doesn't occur for me when saving on the results page, but does occur when saving within the Config Editor. This is more of a nice to have than anything so it wouldn't hold up an approval, but it is strange because looking at the code I'm not sure why it would work in one location and not the other.

julianxcarter (commented):
This is very nice, the download bounce animation is just *chef's kiss*.

I have two comments, not so much change requests because they may be Electron oddities that we're not able to accommodate. Additionally, I saw that the built version of the app does not load the Extract New page by default, and the version ran from node does. I'm not sure how trivial that'd be (I think we'd need to move the contents of `Extract,js` into `App.js` for that to happen?), but if it's a simple ask could we add a default page in the built version as well?

electron/electron.js
Dtphelan1: When you are on the results page, are you saving to the Downloads folder? I know it' sa dumb question, but I just wanna make sure 😄

julianxcarter: No that's reasonable! I am saving to my downloads folder. I'm guessing you're not seeing the same behavior?

Dtphelan1: I'm being honest with you, I didn't test completely and left that as an exercise to the reader 😅. I'll check to see what happens on my machine.

Dtphelan1: Update: It does not in fact work on my machine

julianxcarter (approved):
Looks good to me!

PR #44, opened 2021-11-04
Adding option to start config editor from template

dmendelowitz:
This PR adds a third option to the config editor page to "create from template" which will open the config editor with some fields already filled out. To start, I've added templates for "ICARE" which adds the extractors needed for ICARE and "mCODE" which adds all the extractors (or at least all the ones listed in the MEF's example config). To avoid possible confusion, there is also an alert that will tell the user that some fields have been filled out but others haven't.

To test:
- Go to the config editor
- Click "Create From Template" and pick an option
- See that the correct extractors were added and the config editor works as expected otherwise

### DIFF
--- a/react-app/src/components/ConfigEditor/ConfigEditor.js
+++ b/react-app/src/components/ConfigEditor/ConfigEditor.js
@@ -134,7 +134,7 @@ function ConfigEditor() {
             {showErrorAlert && (
               <Alert variant="info" show={showErrorAlert} onClose={() => setShowErrorAlert(false)} dismissible>
                 <Alert.Heading>Created config from template</Alert.Heading>
-                <p>Extractors have been added for you, but other fields will still need to be filled out</p>
+                <p>Extractors have been added for you, but CSV file paths / URLs and other fields will still need to be added</p>
               </Alert>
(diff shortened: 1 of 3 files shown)


react-app/src/components/ConfigEditor/ConfigEditor.js
```diff
@@ -93,6 +131,12 @@ function ConfigEditor() {
               schema={configSchema}
               closeForm={closeForm}
             />
+            {showErrorAlert && (
+              <Alert variant="info" show={showErrorAlert} onClose={() => setShowErrorAlert(false)} dismissible>
+                <Alert.Heading>Created config from template</Alert.Heading>
+                <p>Extractors have been added for you, but other fields will still need to be filled out</p>
```
julianxcarter, line 137: I think this message kind of implies that the extractors themselves are totally filled out, and that the common extractor args are the only things that need to be added. I would change the message to something along the lines of `Extractors have been added for you, but CSV file paths and other fields will still need to be added`.

julianxcarter (approved):
Looks good to me! Nice work @dmendelowitz

PR #45, opened 2021-11-10
Mask all and string to emails will now convert to UI compatible formats

dmendelowitz:
Now, when loading a config file, `"mask": "all"` will be converted to the full mask array so that it will work in the config editor's UI. Additionally, 'to' emails in the form of a string (e.g. `"email1@test.com, email2@test.com"`) will be converted to an array (`["email1@test.com", "email2@test.com"]`) to be compatible with the extraction UI.

Bonus fixes to make this actually usable: Previously, all constructor args were default set to be hidden in the Extractor objects (even when loading a config with a value present for that arg), so I've changed them to rely on whether or not the constructor arg has a value in the config to decide whether they are hidden or not. Also, the mask and cancer type constructor args didn't actually populate the values from loaded files, so I've also changed those to load properly.

Bonus bonus fix for Windows compatibility: There was a line in `extractor.js` that did a Mac exclusive thing that caused an error on startup on Windows, so I added an if statement to check for the OS first. Mac users please confirm that this still works as it is supposed to.

Testing guidance:
- Make sure `"mask": "all"` in a loaded config file results in all check boxes being checked in the editor
- Make sure a string list of to emails (`"email1@test.com, email2@test.com"`) is properly loaded in the array format of the editor
- Test that loading a config file with any constructor arg results in that constructor arg not being hidden in the extractor in the editor
- Test that mask and cancer type constructor args now properly fill in values from loaded files
- Test that the Mac electron feature that I changed still works on Mac OS


react-app/src/components/ConfigEditor/ExtractorObject.js
```diff
@@ -345,7 +345,7 @@ function ExtractorObject(props) {
                 className="arg-input-width-limit"
               >
                 {arg.options.map((option) => (
-                  <option value={option} key={option}>
+                  <option value={option} key={option} selected={arg[arg.key] === option}>
```
julianxcarter, line 348: The console is throwing a warning on using the `selected` property on the individual options, which is strange since the property exists and seems to be working fine. Instead of using the `selected` property here, could we use the `defaultValue` prop in the parent `From.Select` component? Adding `defaultValue={arg[arg.key]}` should accomplish the same thing and keep the linter happy.

julianxcarter (changes requested):
This looks and works great! One minor comment

julianxcarter (approved):
Looks good!

PR #46, opened 2021-11-15
Select All option for mask field

julianxcarter:
Adds a Select All button for `checkbox` field constructorArgs in `ExtractorObject.js`. This Select All button mirrors the behavior of the Select All button in the file saver found in the extraction results page:
1. Auto-selects when all options are selected individually
2. Clicking the Select All button when all options are not selected will select the remaining options.
3. Clicking the Select All button when all options are selected will deselect all options.


dmendelowitz (approved):
Looks good to me!

PR #47, opened 2021-11-17
Collapse Common Extractor Arguments and Notification Information sections

jafeltra:
This PR focuses on collapsing the Common Extractor Arguments and Notification Information sections in the configuration editor. This lets the focus remain on the more important (and required) Patient Id and Extractor sections. To do this, this PR makes a few key changes:
- Specified an order to the sections that are displayed so the Patient Id CSV File and Extractors come first
- Adds custom components for Common Extractor Arguments and Notification Information. These components wrap the fields in an accordion and then use RJSF's `SchemaField` to render the individiual fields. This was the best middle ground I could find for allowing us to control the surrounding rendering (the accordion) while still utilizing all the functionality RJSF provides for the individual fields (like the input boxes and the component for the To field).
- Reverts back to using the core RJSF `Form` component in order to get the ability to add arbitrary key/value pairs for Request Headers. This functionality doesn't work properly in the bootstrap-4 Form (as evidenced in the [additional properties demo](https://rjsf-team.github.io/react-jsonschema-form/) and documented in this [issue](https://github.com/rjsf-team/react-jsonschema-form/issues/1927)).
  - Reverting back tot he core RJSF `Form` also seems to allow the `classNames` we provide in the uiSchema to work again
  - However, some of the bootstrap-3 classes that the form now uses don't match the ones that our current bootstrap version uses, so I updated a few CSS classes to do their best to mimic the targeted styling in the old class name.
- A few minor changes to the `configSchema`
  - Added the ability to add arbitrary key/value pairs so the UI could allow adding these values
  - Added the "host" and "to" dependencies because the `*` that indicates required was missing on the "To" field when there was "Host" data and vice versa.
 
I removed the `@rjsf/bootstrap-4` since I decided not to use it, but if any of the above changes don't feel that they're worth the styling changes they cause, I can revert and add it back.

This should still work with loading in an existing file, loading a template, and saving a file. No functionality should be lost.

### DIFF
--- a/electron/helpers/configSchema.json
+++ b/electron/helpers/configSchema.json
@@ -8,6 +8,12 @@
       "title": "Patient ID CSV File",
       "type": "string"
     },
+    "commonExtractorArgs": {
+      "$ref": "#/$defs/commonExtractorArgs"
+    },
+    "notificationInfo": {
+      "$ref": "#/$defs/notificationInfo"
+    },
     "extractors": {
       "title": "Extractors",
       "type": "array",
@@ -28,7 +34,10 @@
         },
         "requestHeaders": {
           "title": "Request Headers",
-          "type": "object"
+          "type": "object",
+          "additionalProperties": {
+            "type": "string"
+          }
         }
       }
     },
@@ -66,12 +75,14 @@
       "dependencies": {
         "host": {
           "required": [
+            "host",
             "to"
           ]
         },
         "to": {
           "required": [
-            "host"
+            "host",
+            "to"
           ]
         },
         "from": {
--- a/electron/helpers/configSchema.json
+++ b/electron/helpers/configSchema.json
@@ -8,12 +8,6 @@
       "title": "Patient ID CSV File",
       "type": "string"
     },
-    "commonExtractorArgs": {
-      "$ref": "#/$defs/commonExtractorArgs"
-    },
-    "notificationInfo": {
-      "$ref": "#/$defs/notificationInfo"
-    },
     "extractors": {
       "title": "Extractors",
       "type": "array",
--- a/react-app/package-lock.json
+++ b/react-app/package-lock.json
@@ -8,7 +8,6 @@
       "name": "app",
       "version": "0.1.0",
       "dependencies": {
-        "@rjsf/bootstrap-4": "^3.1.0",
         "@rjsf/core": "^3.0.0",
         "@testing-library/jest-dom": "^5.11.4",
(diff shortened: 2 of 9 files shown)


dmendelowitz (commented):
Overall, this looks really good! The only thing that stood out to me is the weird "newKey Key" labelling on the request headers which could be confusing to users. Maybe that's fine if we're assuming this optional field is only gonna be used by people who know what they're doing? What do you think?

jafeltra:
>The only thing that stood out to me is the weird "newKey Key" labelling on the request headers which could be confusing to users. Maybe that's fine if we're assuming this optional field is only gonna be used by people who know what they're doing? What do you think?

I agree that this feels pretty jarring to me, and it always throws me off when the labels above the fields update as you type in the Key field. However, it looks like the "newKey" (and the index appended to it) are hardcoded as the default new key name in RJSF 😕 I hadn't confirmed that that was the case until now, but I don't see a way around it. Until now, I was thinking that since this adds the ability to build out Request Headers where we couldn't at all previously made this change worthwhile. But if you were also confused by it, maybe it really isn't worth it.

> Maybe that's fine if we're assuming this optional field is only gonna be used by people who know what they're doing? 

To this point, I think it's probably fairly unlikely that users use this field, at least initially, because I believe the only time we need to set Request Headers is when we actually make web service calls, which we don't support in the UI yet. This could also be another argument to remove the Request Headers section entirely, but I just hadn't convinced myself we wanted to do that yet.

None of this really answered your question. Do you think the "newKey" thing is too confusing and we should get rid of it entirely? Or do you think since it's likely it's never used, we keep the functionality in case anyone does in fact use it?

jafeltra:
Incoming 🎢 :

I discovered that you can add the following `additionalProperties` to the uiSchema like:

```
requestHeaders: {
      classNames: 'additional-properties-width input-array-label',
      additionalProperties: {
        'ui:title': 'Request Header',
      },
    }
```

In theory, the `ui:title` should allow the new key/value pairs to be labeled a little more intuitively like so:

image

However, once you edit one of those values, the user provided title from the schema doesn't seem to stay as the preference, and it ends up looking like this:

image

I feel like this is even more confusing that it is without any sort of title, so I'll hold off on adding this. We can decide later how to handle this when Request Headers is more likely to be used when/if we add Web Service extractors to the UI.

With that 🎢 , I think there isn't anything else to address here. Let me know if you notice anything else!

dmendelowitz (approved):

PR #48, opened 2021-12-01
Use HashRouter and update Links

jafeltra:
This PR switches from BrowserRouter to HashRouter. This makes it so when you build the app and launch it, it opens on the Extract New page. There was also a warning in the console saying "Warning: Hash history cannot PUSH the same path; a new entry will not be added to the history stack" when clicking on the navigation bar links, so I added the `replace` prop to the NavLinks in the NavBar. If anyone knows why we were using `useHistory` vs the `Link` and I shouldn't have changed this, please let me know!

While looking into that, I also realized we could be using the `Link` component from react router instead of `useHistory`, so I updated the `LinkButton` component we created and used it in another location as well.

I also ran prettier since a few files were out of date, so all changes in `ConfigEditor` and `ExtractorObject` are solely from prettier. I didn't make any meaningful changes there.


react-app/src/components/CommonComponents/Navbar.js
```diff
@@ -6,16 +6,16 @@ function Navbar() {
   return (
     <NavBar collapseOnSelect fixed="top" variant="dark" expand="lg" className="navbar">
       <Container fluid>
-        <NavBar.Brand as={NavLink} to="/" className="navbar-brand">
+        <NavBar.Brand as={NavLink} to="/" className="navbar-brand" replace>
```
Dtphelan1, line 9: DP1: What is the replace prop doing here?

Dtphelan1: Ditto for all the other Nav Links

jafeltra: I don't think it's completely necessary, but I added it because it prevents a warning in the console in a way that I don't think harms anything. With HashRouter, I was getting a warning saying "Hash history cannot PUSH the same path; a new entry will not be added to the history stack" and the `replace` prop makes it so that clicking the link replaces the route rather than push a new one (documentation [here](https://v5.reactrouter.com/web/api/Link/replace-bool)). Since there isn't any navigation we can do in this app besides clicking the routes themselves or a button that directly reroutes you somewhere, I figured it was safe to replace the entry (I don't know a case where it wouldn't be safe, so this is kind of an invented case and a best guess at what could go wrong). From looking at some issue threads, it doesn't look like that warning is harmful at all and is mostly intended to help in development (so you have an idea of why nothing might be happening), so I can definitely take out the `replace` flags, but there will be new warnings that pop up in the console (in development) occasionally.

Dtphelan1: Very informative. No complaints. Full send 🚀

Dtphelan1 (approved):
Build works and changes look good!

PR #49, opened 2021-12-02
Load templates staticly, not dynamically

Dtphelan1:
Address STEAM-795 by loading the mcode/icare templates in memory, avoiding the need for the backend to dynamically load these files. Since the loadFile functions are used in other ways, we didn't need to remove any electron-preload functions.


jafeltra (approved):
Looks good to me! The build works as expected, and it still works in development!

PR #50, opened 2021-12-08
Adding start script to check for OS and call appropriate commands

dmendelowitz:
This PR adds a new script, `start.js`, that is called by `npm start` and checks for operating system, then calls the appropriate commands to start the app on that OS. The app should run exactly the same as before on Linux/Mac, but will now run properly in Windows when called with `npm start`


jafeltra (approved):
This looks good to me! And it started on the VM for me!

PR #51, opened 2021-12-15
Disallowing relative filepaths in loaded config files

dmendelowitz:
This PR adds a check for relative paths after the config file is loaded and before extraction starts, if there are relative paths in the patientIdCsvPath or any extractor filePath fields, an error will occur telling the user that they cannot use relative paths for these fields. To test: load a config file with a relative path in any extractor object or in the patientIdCsvPath field and you should see an error upon submitting.

I did consider various ways to immediately link the user to the config editor to fix this but then remembered that we had STEAM-716 (Design linking the configuration file editing with the Extract New page) in the backlog, so I went with a simpler solution for now, and I'll just note here that converting a relative path config file to an absolute path config file would be a situation where linking the Extract New page and config editor would be very useful.

### DIFF
--- a/electron/extraction.js
+++ b/electron/extraction.js
@@ -1,3 +1,4 @@
+const { isAbsolute } = require('path');
 const { getConfig, logger, mcodeApp, MCODEClient } = require('mcode-extraction-framework');
 
 async function runExtraction(fromDate, toDate, configFilepath, runLogFilepath, debug, allEntries) {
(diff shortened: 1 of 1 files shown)


Dtphelan1 (approved):
I think this looks great. Works on my machine plenty well, and I agree that we don't need to link people to the config editor as a part of this task. No complaints here, and happy holidays 👍

PR #52, opened 2021-12-17
Update build process

jafeltra:
This PR allows Mac users to build the application for Mac and Windows, and it allows Windows users to build for Windows (after removing the `-m` flag from the build script). You should be able to build the application and open the relevant built app on your computer.

I also removed some of the old electron-forge related scripts, although I'm not sure I know what all of them were used for. If anyone knows of one I should leave or replace with an electron-builder version, let me know.

The new build approach also resolves [STEAM-792](https://jira.mitre.org/browse/STEAM-792). As long as reviewers see that the app logo appears, we can close that task as well.


dmendelowitz (commented):
This still works on Windows! My only comment is it might be good to add another script like "build:windows" or something that does the same thing but with just "-w" instead of "-mw" so windows users don't have to remember to change the script every time

jafeltra:
> My only comment is it might be good to add another script like "build:windows" or something that does the same thing but with just "-w" instead of "-mw" so windows users don't have to remember to change the script every time

Oh that's a great idea! I'll add that!

jafeltra:
@dmendelowitz added the separate script in 90af73f!

dmendelowitz (approved):

Dtphelan1 (approved):
Can confirm this build works on a mac!  Looks good to me Jules 😊

